PR #1, opened 2020-11-10
Feature/gh pages

aloiscochard:


package.json
```diff
@@ -34,7 +34,9 @@
     "start": "react-scripts start",
     "build": "react-scripts build",
     "test": "react-scripts test",
-    "eject": "react-scripts eject"
+    "eject": "react-scripts eject",
+    "predeploy": "npm run build",
+    "deploy": "gh-pages -d build"
```
polarker: Do we have to define these two scripts: `predeploy` and `deploy` ? I don't like this, as it breaks the consistency of the code

tdroxler (approved):

PR #3, opened 2020-11-20
Feature/default address

aloiscochard:


.github/workflows/gh-pages.yml
```diff
@@ -14,7 +14,7 @@ jobs:
       - name: Install and Build 🔧 
         run: |
           npm install
-          REACT_APP_ALEPHIUM_HOST='3.121.42.238' npm run build
+          REACT_APP_ALEPHIUM_HOST='https://explorer-backend.alephium.org:9443' npm run build
```
polarker, line 17: Let's use `REACT_APP_BACKEND_URL` to make it more clear

PR #5, opened 2020-11-25
Feature/gh 404

aloiscochard:


public/404.html
```diff
@@ -0,0 +1,40 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="utf-8">
+    <title>alephium-explorer</title>
```
polarker, line 5: Let's use `Alephium Explorer` instead, which is more common style

public/index.html
```diff
@@ -26,6 +26,31 @@
     -->
     <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Roboto:300,400,500,700&display=swap" />
     <title>alephium-explorer</title>
```
polarker, line 28: same here

.github/workflows/gh-pages.yml
```diff
@@ -1,7 +1,5 @@
 name: Build and Deploy
-on:
-  push:
-    branches: [ master ]
+on: [push]
```
polarker, line 2: we need to revert to push-master

PR #8, opened 2020-12-01
Use hash router

polarker:


src/index.tsx
```diff
@@ -17,7 +17,7 @@
 import React, { useState, useEffect, useRef, useCallback } from 'react';
 import ReactDOM from 'react-dom';
 import './index.css';
-import { BrowserRouter as Router, Redirect, Route, useLocation } from 'react-router-dom'
+import { HashRouter as Router, Redirect, Route, useLocation } from 'react-router-dom'
```
tdroxler, line 20: what's the difference?

polarker: this works better with Github pages: https://create-react-app.dev/docs/deployment/#notes-on-client-side-routing
the url will be like `https://devnet.alephium.org/#/blocks` with a hash character in between

src/utils/util.tsx
```diff
@@ -104,3 +104,7 @@ export function useInterval(callback: () => any, delay: number, shouldPause = fa
     }
   }, [delay, shouldPause]);
 }
+
+export function smartHash(hash: string) {
+  return hash.substring(0, 8) + '...' + hash.substring(hash.length - 8)
```
tdroxler, line 102: maybe we should have something like 
```
if (hash.length < 10)
   return hash
else
  return hash.substring(0, 8) + '...' + hash.substring(hash.length - 8)
```
we currently don't use this function on short hash, but nothing prevent us to do it in the future

src/components/Amount.tsx
```diff
@@ -22,6 +22,7 @@ interface AmountProps {
 }
 
 const Amount: FC<AmountProps> = ({ value, className }) => {
+  console.log('value' + value?.toString())
```
tdroxler, line 25: is it a debug leftover or you really want it?

tdroxler:
great to start adding tests :+1:

tdroxler (approved):
minor comments that you can tackle or not, otherwise lgtm

src/components/Amount.tsx
polarker: Nice catch!

PR #18, opened 2021-07-16
Pagination system for lists (TX)

mvaivre:
Add pagination system in all list, page state stored in the URL.
Split loading between entity info vs. transactions.


src/sections/TransactionInfoSection.tsx
```diff
@@ -106,6 +107,18 @@ const TransactionInfoSection = () => {
                     ))}
                   </td>
                 </tr>
+                <tr>
+                  <td>Start Gas</td>
+                  <td>
+                    <Amount value={JSBI.BigInt(txInfo.data.startGas)} />
```
tdroxler, line 113: start gas isn't a big int, it's a regular int

tdroxler (approved):

PR #21, opened 2021-08-30
Show TX status, with polling mechanism + layout improvements

mvaivre:
PR content:

- Show TX status in TransactionInfoSection
- Polling until the TX is "confirmed"
- New "Copy to clipboard" button near important info (addresses etc.)
- More compact UI, better experience on small screen
- Tables are now scrollable on mobile
- Sidebar open/close mechanism is clearer
- startGas -> gasAmount


tdroxler (approved):
Reviewed and tested (also the unconfirmed tx page) LGTM

PR #22, opened 2021-09-28
UX fine tuning

mvaivre:
- Better contrast in light theme
- Stack "body only" tables vertically on mobile
- Fix address / hash validation
- Update deps & some UI fixes and improvements


src/utils/strings.ts
```diff
@@ -0,0 +1,22 @@
...
+//
+// The library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with the library. If not, see <http://www.gnu.org/licenses/>.
+
+export const createRandomId = () => Math.random().toString(36).substring(7)
```
polarker, line 17: The idea is to return a random string of 29 characters? How will this function be used?

src/components/Sidebar.tsx
```diff
@@ -14,7 +14,7 @@
 // You should have received a copy of the GNU Lesser General Public License
 // along with the library. If not, see <http://www.gnu.org/licenses/>.
 
-import React, { useContext, useEffect, useRef } from 'react'
+import { useContext, useEffect, useRef } from 'react'
```
polarker, line 17: Is there any lint tool to force clean dependencies ? Scalastyle did a great job for us

nop33: Probably this [plugin](https://www.npmjs.com/package/eslint-plugin-unused-imports) of [eslint](https://eslint.org/) should do it.

PR #23, opened 2021-10-19
Add audit to test CI, move react-scripts to devDependencies

mvaivre:
Even though the usefulness of `npm audit` is debated for front-end apps, let's use it to check our production dependencies.

Interesting reads: 
https://github.com/facebook/create-react-app/issues/11174
https://overreacted.io/npm-audit-broken-by-design/

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2,9 +2,9 @@ name: Test
 
 on:
   push:
-    branches: [ master ]
+    branches: [master]
   pull_request:
-    branches: [ master ]
+    branches: [master]
 
 jobs:
   test:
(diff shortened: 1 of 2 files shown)


polarker:
> Even though the usefulness of `npm audit` is debated for front-end apps, let's use it to check our production dependencies.
> 
> Interesting reads: [facebook/create-react-app#11174](https://github.com/facebook/create-react-app/issues/11174) https://overreacted.io/npm-audit-broken-by-design/

I don't think that's the problem for `npm audit`. `npm audit` cannot know which dependency is a build tool or not. Instead `create-react-app` should fix the problem if they cannot pas `npm audit`

PR #25, opened 2021-11-02
Network switch & explicit environment variables setting

mvaivre:
New switch in the sidebar, currently allowing switching from mainnet to testnet explorers. 
Backend URL is now to be set explicitly or it breaks (following a discussion with @killerwhile).

Do not merge before confirmation about production URL used for our mainnet explorer.


.env.placeholder
```diff
@@ -1 +1,4 @@
-REACT_APP_BACKEND_URL=https://testnet-backend.alephium.org
\ No newline at end of file
+REACT_APP_BACKEND_URL=https://testnet-backend.alephium.org
+
+# Network: testnet / mainnet
+REACT_APP_NETWORK=testnet
```
tdroxler: should we now use mainnet as default?

src/index.tsx
```diff
@@ -42,8 +42,18 @@ import { useHistory } from 'react-router-dom'
 import { isElectron } from './utils/misc'
 import { ScrollToTop } from './utils/routing'
 
+const networkTypes = ['testnet', 'mainnet'] as const
+export type NetworkType = typeof networkTypes[number]
+
+interface SnackbarMessage {
```
tdroxler, line 48: Can I order a cheeseburger?

tdroxler (approved):
minor comment, LGTM

PR #29, opened 2021-11-22
Ability to generate and show address' QR code

### LINKED ISSUE #28: Add QR code in Address page
@polto suggested:

> I think a QRcode of the address would be appreciated on the address page such as https://explorer.alephium.org/#/addresses/1CtfRFQ4oLMfk9yyZTXiMEPSEf1S8wQMSysQvsS1pVk7Z

mvaivre:
Show a little QR code icon next to the address.
QR code is shown in a tooltip, reusing our existing components. 
Tested on desktop and on mobile.

Other modifications:

- Cleanup Buttons' folders hierarchy
- Switch from using feather icons to lucide icons (a fork with more icons, better maintained).

Fixes #28


src/components/PageSwitch.tsx
```diff
@@ -1,9 +1,9 @@
 import { useCallback, useEffect, useMemo } from 'react'
-import { ChevronRight, ChevronLeft } from 'react-feather'
+import { ChevronRight, ChevronLeft } from 'lucide-react'
 import { useHistory } from 'react-router-dom'
 import styled from 'styled-components'
 import useQueryParams from '../hooks/useQueryParams'
-import { TextButton } from './Buttons'
+import { TextButton } from './Buttons/TextButton'
```
nop33, line 6: I know this was not changed in this PR, but `Buttons/TextButton` should be a default export instead of a named export, similarly to the `ClipboardButton` and `QRCodeButton`, so that it can be imported like so:

```js
import TextButton from './Buttons/TextButton'
```

nop33 (commented):
Just one comment, otherwise LGTM 👏  Tested on my machine on Safari, Brave, and Firefox, seems to be working fine.

src/components/PageSwitch.tsx
mvaivre: It was changed in this PR actually. Nice catch.

nop33 (approved):

mvaivre:
Had to switch to another QR code package (same as in our wallet) to avoid an unused peer-dep (used only on React Native) which introduced a vulnerability.

nop33 (approved):
Works well! LGTM

PR #40, opened 2021-12-22
Show locked balance

### LINKED ISSUE #36: Show amount of locked tokens
Goal is to show both the total balance + the portion of tokens that is currently locked.
We don't show locking periods for now. 

cc/ @tdroxler

nop33:
Needs to be properly tested once alephium-js schema gets updated: https://github.com/alephium/alephium-js/pull/34

Closes #36

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,6 +7,7 @@
     "npm": ">=7.24.2"
   },
   "dependencies": {
+    "alephium-js": "1.0.5",
     "dayjs": "^1.10.7",
     "framer-motion": "^4.1.17",
     "jsbi": "^3.2.4",
--- a/src/utils/api.ts
+++ b/src/utils/api.ts
@@ -1,0 +1,30 @@
+// Copyright 2018 - 2021 The Alephium Authors
+// This file is part of the alephium project.
+//
+// The library is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Lesser General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// The library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with the library. If not, see <http://www.gnu.org/licenses/>.
+
+import { tryGetString } from './types'
+
+interface APIError {
+  error: { detail: string }
+}
+
+export const isHTTPError = (e: unknown): e is APIError => {
+  return (e as APIError).error !== undefined
+}
+
+export const getHumanReadableError = (e: unknown, defaultErrorMessage: string) => {
+  const stringifiedError = tryGetString(e)
+  return isHTTPError(e) ? e.error.detail : stringifiedError || defaultErrorMessage
+}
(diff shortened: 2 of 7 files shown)


tdroxler:
Can we have a screenshot of the result :) you can check a genesis address with locked balance from https://github.com/alephium/dev-alephium/blob/master/flow/src/main/resources/mainnet_genesis.conf

nop33:
image

src/utils/api.ts
```diff
@@ -0,0 +1,30 @@
+// Copyright 2018 - 2021 The Alephium Authors
```
nop33, line 1: @mvaivre this file and the next one are a good example of duplicated code between our projects.

I copied them from the alephium-wallet project. Since I started using alephium-js in this project, our error handling needs to adapt to the responses of the alephium-js API classes. This is how the wallet handles it and I think this is how the explorer should handle it as well.

The question is: How do we handle shared logic between our projects? Is it too early to start thinking of shared codebases? Maybe we could add these 2 files in the alephium-js?

Tbh, it makes me feel a bit dirty copying the same files amongst projects, but maybe it's the best approach for the moment. Happy to discuss.

package.json
```diff
@@ -7,6 +7,7 @@
     "npm": ">=7.24.2"
   },
   "dependencies": {
+    "alephium-js": "1.0.5",
```
nop33, line 10: @mvaivre, since I had to update the API schema for this feature, I took the opportunity to integrate alephium-js slowly (one use-case at the time).

src/utils/api.ts
nop33: Note to myself: I'll probably need to update the code to reflect this bug fix: https://github.com/alephium/alephium-wallet/pull/94

mvaivre: Everything started in the explorer before going into the wallet. As you saw, some things changed along the way (error format etc.), everything was constantly moving so yeah - it was too soon to put everything in a shared codebase.

Now might be the time to go back to centralising. 
Error management and amount display could be a good start.

It would be good to discuss how with structure `alephium-js` in future iterations though. For now, it was essentially focused on api methods (flat hierarchy). Now we're gonna start adding UI helpers. Distinction should be made clear.

package.json
mvaivre: Good stuff!

mvaivre:
For clarity sake, maybe we could rename `Balance` into `Total balance`. WDYT?

nop33:
> For clarity sake, maybe we could rename `Balance` into `Total balance`. WDYT?

Totally agreed.

src/utils/api.ts
nop33: Yeah, if we want to "sell" the alephium-js project as the SDK for building smart contracts on Alephium, then UI helpers definitely have no place there.

What we essentially see emerging is common functionality (as you correctly said, error formating, amount display, etc) amongst our React apps (explorer, wallet). We will build more React apps, including the dapps.

Maybe adding a couple UI-related shared functions to alephium-js is not that bad of an idea to start with. The more we find common functionality, the more ideas will be generated as to how to handle separation of concerns.

mvaivre: Constants should also be in alephium-js. I'm about to add some in the wallet (minGas etc.). I don't want to add them to the SDK right away, as we need to discuss this together first. 

But to summarise quickly, I see:

- API methods
- Formatting helpers (numbers, amounts, errors etc.)
- Constants

.. being thrown in alephium-js. 

I'll create a task there to continue the discussion 🚀

mvaivre:
LGTM. This can be rebased / merged and deployed :)

nop33:
@mvaivre I rebased, I was about to merge, but realized there's no "official" GitHub approval yet, so I'll wait :D

mvaivre (approved):
There you go :))

PR #49, opened 2022-01-17
Move number utils to alephium-js and display notification if browser is too old

nop33:
We need to release a new version of alephium-js to remove the "DRAFT" status. Tested locally using `npm link`.

We have to check that BigInt is supported. If not, we display a notification to the user that their browser version is too old: https://caniuse.com/?search=bigint

Tested with Safari <= 13

https://user-images.githubusercontent.com/1579899/149785479-ec3e613c-adc3-45e8-885d-2a117e04d87b.mp4


nop33:
@mvaivre "DRAFT" status removed. Ready for review.

src/index.tsx
```diff
@@ -359,9 +360,23 @@ const SnackbarPopup = styled(motion.div)`
   }
 `
 
+let browserIsOld = false
+
+try {
+  BigInt(1)
+} catch (e) {
+  browserIsOld = true
```
mvaivre, line 368: Nice one!

src/sections/BlockInfoSection.tsx
```diff
@@ -210,19 +210,18 @@ const TransactionRow: FC<TransactionRowProps> = ({ transaction }) => {
...
         <td>
-          {t.inputs.length} address{t.inputs.length > 1 ? 'es' : ''}
+          {t.inputs ? t.inputs.length : 0} {t.inputs && t.inputs.length === 1 ? 'address' : 'addresses'}
         </td>
         <td>
           <ArrowRight size={15} />
         </td>
         <td>
-          {t.outputs.length} address{t.outputs.length > 1 ? 'es' : ''}
+          {t.outputs ? t.outputs.length : 0} {t.outputs && t.outputs.length === 1 ? 'address' : 'address'}
```
mvaivre, line 219: address**es**

src/sections/TransactionInfoSection.tsx
```diff
@@ -200,4 +198,6 @@ const TransactionInfoSection = () => {
   )
 }
 
+const isTxConfirmed = (tx: TransactionLike): tx is Transaction => true
```
mvaivre, line 201: Shouldn't the type assertion logic be defined in the function?
You always return `true` in that case. Not sure Typescript can actually compare types at runtime?

Cf: https://stackoverflow.com/questions/44078205/how-to-check-the-object-type-on-runtime-in-typescript

src/index.tsx
mvaivre: No need to define e as a param though (eslint wouldn't be happy I believe?)

src/sections/BlockInfoSection.tsx
nop33: - 0 addresses
- 1 address
- 2 addresses

nop33: - _"I have 0 dollars in my wallet"_
- _"I have 1 dollar in my wallet"_
- _"I have 2 dollars in my wallet"_

src/sections/TransactionInfoSection.tsx
nop33: I implemented the above code after reading the same sources that you posted. The link to the user-defined type guard is a depricated page on the TypeScript documentation. The "updated" way to implement the same logic, is the one above.

src/sections/BlockInfoSection.tsx
mvaivre: mhh yes? Please look at your code.

nop33: oops-a-daisy, sometimes I speak before I think, sorry!

src/index.tsx
nop33: yep, you believe correct.

image

src/sections/TransactionInfoSection.tsx
mvaivre: [Here's ](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates)the up-to-date page. 

I'm confused. It looks to me that `isTxConfirmed` will always return true, as you don't return a type predicate, ie. a piece of logic to say when the input must be considered as a `Transaction` (like checking for a certain field presence etc.)

nop33: Interesting, now I get redirected to a different page than I was before when clicking on the deprecation warning message!

nop33: I can't find the source anymore. Besides, I was also baffled by this quite a lot. Initially I started by checking for the presence of `blockHash`, but somehow I ended up there. Updating the code to this:

```tsx
const isTxConfirmed = (tx: TransactionLike): tx is Transaction => {
  return (tx as Transaction).blockHash !== undefined
}
```

I've [verified](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgCpTiAzoswD2IyA3gLABQyVyAFnFjQFzJZhSgDmA3BdcgEYAbfAgDWACXpMWbTj0rU8AWwis4SgA7MQAVyX9o8gL4VQkWIhQBVEAkIxgUFQBNUADxK9qdBs1bsQbi8qBDpQADEofCVtPQMoeT5QuFBUfFj9QwoTcmcIBEE4KBQwAE8NFHRMHAQ8QgAZYFEUAF40DGxcAiIAH2QbOxAHJwhXN3kKQdZkYCx3AGF7RxdkNoAKMDdmKs7a7sbmgEpmTZmsduquwlWAPk8FYrAdKCINj3oL3bqQQ4A6IREEikyAAhC02joQHkHCBRtkKJNCNMlKUdjVvtsOujuqsSD5pAByNwEgA0AmEYkkvmQBNKpOQylUYHUWmQAEYclMwMgUQMliNXFiriBmHyhstRu5ccR8cwifTkhEojF2WTFSA0swOQjyFN8IIIL9hBw1rMFvyXGsUWjhYdDojsPrDcbTXM3ItxQKraUxcMXDa9oQ7UA) that this works.

polarker:
I remember big integer was not well supported on mobile phone, that's why we used `jsbi`. Is this situation improved?

nop33:
> I remember big integer was not well supported on mobile phone, that's why we used jsbi. Is this situation improved?

@mvaivre and I discussed this. BigInt is not supported in Safari 13 and prior (https://caniuse.com/?search=bigint). Since we don't have analytics on explorer, we checked the analytics of our website. There we can see that around `3.5%` of our website visitors are using unsupported browser versions. Thus, we decided to detect the browser version and display a message to those users that they need to upgrade their browser (see video). The alternative would be to use JSBI in alephium-js, and adapt all the code of the wallet (which uses BigInt) to use JSBI instead.

image

src/sections/TransactionInfoSection.tsx
mvaivre: I'm actually relieved. This makes sense now 🙃

mvaivre (approved):

polarker:
> > I remember big integer was not well supported on mobile phone, that's why we used jsbi. Is this situation improved?
> 
> @mvaivre and I discussed this. BigInt is not supported in Safari 13 and prior (https://caniuse.com/?search=bigint). Since we don't have analytics on explorer, we checked the analytics of our website. There we can see that around `3.5%` of our website visitors are using unsupported browser versions. Thus, we decided to detect the browser version and display a message to those users that they need to upgrade their browser (see video). The alternative would be to use JSBI in alephium-js, and adapt all the code of the wallet (which uses BigInt) to use JSBI instead.
> 
> image

Thanks for the details! LGTM

PR #50, opened 2022-01-17
Display info message about orphan blocks

### LINKED ISSUE #43: Orphan block display
The explorer could show orphan blocks, for example: https://explorer.alephium.org/#/blocks/00000000004f694568fc40867c8a63da2ee76827e2ef2d9ae5f5ff49cc16e187

But the mining rewards of the tx is zero, as the block&tx is not on mainchain

nop33:
Closes #43

Looks like this: 

image

### DIFF
--- a/src/sections/BlockInfoSection.tsx
+++ b/src/sections/BlockInfoSection.tsx
@@ -149,30 +149,39 @@ const BlockInfoSection = () => {
                 </TableBody>
               </Table>
 
-              <SecondaryTitle>Transactions</SecondaryTitle>
-              {!txLoading && txList && txList.data && txList.status === 200 ? (
+              {blockInfo?.data?.mainChain ? (
                 <>
-                  <Table main hasDetails scrollable>
-                    <TableHeader
-                      headerTitles={['', 'Hash', 'Inputs', '', 'Outputs', 'Total amount', '']}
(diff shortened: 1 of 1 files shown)


src/sections/BlockInfoSection.tsx
```diff
@@ -149,30 +149,39 @@ const BlockInfoSection = () => {
...
+                  <SecondaryTitle>Orphan block</SecondaryTitle>
+                  <div>It appears that this block is not part of the main chain.</div>
+                </>
               )}
             </>
           ) : (
             <InlineErrorMessage message={blockInfo?.detail} code={blockInfo?.status} />
           )}
-          {txList && txList.data && blockInfo?.data?.txNumber && (
+          {txList && txList.data && blockInfo?.data?.txNumber !== undefined && blockInfo.data.txNumber > 0 && (
```
tdroxler, line 184: great, like this if in the future we return the txs for orphan blocks, it will directly work :+1:

tdroxler (approved):

src/sections/BlockInfoSection.tsx
nop33: Not exactly, this was fixing a different bug I noticed where the number `0` was displayed on the screen. This fixes it.

When in the future we return txs for orphan blocks, we'd need to change the condition in line `152`

😇

mvaivre (approved):

PR #51, opened 2022-01-20
Timestamp format switch

### LINKED ISSUE #48: Give option to show the exact timestamps
- Users could switch the display mode timestamps by clicking on one. 
- We display a little tip on hover ("Click to see the exact time") when in simplified timestamp mode.
- The exact time should specify the current timezone (would fix #34).

mvaivre:
Fixes #48

Note1: the tooltip is not refreshed when clicking on the timestamp to change the format. This is a minor issue that I mitigated by hiding the tooltip on click. Calling ReactTooltip.rebuild()` didn't help. We could have forced parent remounting, but didn't want to make this overcomplicated.

Note2: Imports order is wrong in several places + some file headers are still missing. I'll add the proper lint configuration & plugins in a separate PR.

### DIFF
--- a/src/components/Timestamp.tsx
+++ b/src/components/Timestamp.tsx
@@ -1,28 +1,59 @@
+// Copyright 2018 The Alephium Authors
+// This file is part of the alephium project.
+//
+// The library is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Lesser General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// The library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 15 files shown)


src/components/Timestamp.tsx
```diff
@@ -0,0 +1,61 @@
...
+      <span
+        onClick={handleTimestampClick}
+        data-tip={`
+        ${isHighPrecision ? lowPrecisionTimestamp : highPrecisionTimestamp}
+        ${!forceHighPrecision ? '<br/>(Click to change format)' : ''}`}
+        data-multiline
+      >
+        {isHighPrecision ? highPrecisionTimestamp : lowPrecisionTimestamp}
+      </span>
+    </>
```
nop33, line 57: Since this is only 1 element, is the fragment needed?

mvaivre: Removed

nop33 (approved):

PR #53, opened 2022-01-21
Better linting

### LINKED ISSUE #52: Dev UX: add auto-lint for file headers and imports order

mvaivre:
Configure eslint and fix issues for file header and imports.

Fixes #52


package.json
```diff
@@ -47,8 +50,9 @@
     "build": "react-scripts build",
     "test": "react-scripts test",
     "eject": "react-scripts eject",
-    "lint": "eslint . --ext .ts,.tsx",
-    "lint:fix": "eslint . --fix --ext .ts,.tsx"
+    "lint": "eslint . --ext .ts,.tsx --max-warnings=0",
```
nop33, line 53: Why `max-warnings=0`? We don't want warnings to trigger non 0 exit code?

mvaivre: It's something that I added in the wallet project when we had the issue with the  css variable color animation (if I'm not mistaken). Eventually I'm not sure warnings should make the CI fail, but let's be strict for now. I'll remove it.

mvaivre: Done

nop33 (approved):

PR #61, opened 2022-03-16
Statistics

lf94:


lf94:
Blocked by https://github.com/alephium/explorer-backend/issues/169

src/components/StatisticBlock.tsx
```diff
@@ -0,0 +1,72 @@
...
+  border: 1px solid ${({ theme }) => theme.borderSecondary};
+  box-shadow: ${({ theme }) => theme.shadowPrimary};
+  border-radius: 7px;
+  display: flex;
+  width: 100%;
+  flex-direction: column;
+  padding: 2rem;
+`
+
+const Title = styled.div`
```
mvaivre, line 35: So, some "conventions" we've agreed on with @nop33 (conventions, yay! :D) :
- We always put the styled components after the main exported react component (which is `StatisticBlock` here)
- We put the types / interface first.

So we always have 1) Prop types 2) exported component 3) styling.

src/components/StatisticBlock.tsx
```diff
@@ -0,0 +1,72 @@
...
+`
+
+interface Props {
+  title: string
+  primary: ReactNode
+  secondary: ReactNode
+  isLoading: boolean
+}
+
+const StatisticBlock: FC<Props> = ({ title, primary, secondary, isLoading }) => {
```
mvaivre, line 57: As we don't use `children`, we don't need to use `FC`. We can just go like: 

```ts
const StatisticBlock = ({ title, primary, secondary, isLoading }: Props) => 
```

And as this is a "dump component" (presentational only, no logic), we don't need the brackets and the return statement :)

src/components/StatisticBlock.tsx
lf94: o7 roger that

src/components/StatisticBlock.tsx
lf94: Habit to just make everything FC<>, even if a dumb component, will remove :+1:

lf94:
To test the code you'll have to check out and run https://github.com/alephium/explorer-backend/tree/feature/nb-of-tx-endpoint, and generate your own alephium-js package to make use of the new OpenAPI spec.

src/components/StatisticBlock.tsx
lf94: Gah, I still need to do this.

I would still use `FC` but you're right about removing the `return` - it's 2 characters for a more accurate type. If you feel strongly about omitting it I'll do it though, I just don't see any good reason other than "it adds more semantics to the code", which is a good reason I guess but not one I agree with :slightly_smiling_face:

nop33: I haven't seen any benefit of having a `FC` over a normal function so far. So we decided to drop it when not needed and save one line of imports.

README.md
```diff
@@ -0,0 +1,37 @@
+# Alephium Explorer - Front-end
+
+A blockchain explorer front-end for Alephium.
+
+## Features
+
+* Switch between main and test networks
```
nop33, line 7: I think an `npm run format` would update this file.

package.json
```diff
@@ -7,7 +7,7 @@
     "npm": ">=7.24.2"
   },
   "dependencies": {
-    "alephium-js": "1.1.0",
+    "alephium-js": "file:../alephium-js",
```
nop33, line 10: This should be changed before merging. Also, using `npm link` removes the need to change this.

src/components/StatisticBlock.tsx
```diff
@@ -0,0 +1,80 @@
...
+      {!isLoading && (
+        <Block>
+          <Title>{title}</Title>
+          <Primary>{primary}</Primary>
+          <Secondary>{secondary}</Secondary>
+        </Block>
+      )}
+    </>
+  )
+}
```
nop33, line 48: Nit: I would personally do it this way:
1. Remove `FC` since we don't need `children` (same for other components in other files)
2. Remove `return` keyword since it's a dumb component
3. Use a ternary operator to avoid using a fragment

```suggestion
const StatisticBlock = ({ title, primary, secondary, isLoading }: Props) =>
  isLoading ? (
    <BlockCenteredContent>
      <LoadingSpinner />
    </BlockCenteredContent>
  ) : (
    <Block>
      <Title>{title}</Title>
      <Primary>{primary}</Primary>
      <Secondary>{secondary}</Secondary>
    </Block>
  )
```

src/components/Statistics.tsx
```diff
@@ -0,0 +1,158 @@
...
+            abbreviateValue(values.reduce((acc: number, { value }) => acc + value, 0.0) / values.length, SUFFICES_TIME)
+          )
+        )
+    ]).then(() => setIsLoading(false))
+  }, [refresh, client])
+
+  return (
+    <Block>
+      <StatisticBlock
+        title={'Hashrate'}
```
nop33, line 89: This should work as well, right? (also below)

```suggestion
        title="Hashrate"
```

src/components/StatisticBlock.tsx
lf94: Fair enough! I'll omit it.

----

The comparison I see is you wouldn't just leave a function untyped normally, because it's bad type safety: `const f = (s: string): string => 'takes and returns a string';`
So I don't (personally) compromise when having a function which returns an element either: `const F = (p: Props):ReactElement => <></>;` - `FC` is just shorthand for this. You never see this as an issue though because TypeScript is inferring that `<></>` is an `Element` so it deduces that `F` returns an `Element`, but still, it's not good because what if you made a mistake in the function which returns a `string` or a type other than `Element`?

src/components/Statistics.tsx
```diff
@@ -0,0 +1,158 @@
...
+          <span>
+            <TextPrimary>{currentSupplyPercentage}%</TextPrimary> is circulating
+          </span>
+        }
+        isLoading={isLoading}
+      />
+      <StatisticBlock
+        title={'Transactions'}
+        primary={abbreviateValue(parseInt(transactions), SUFFICES_QUANTITY)}
+        secondary={<div>Total</div>}
```
nop33, line 122: Are these `div`s really needed? Their parent seems to be just another `div`. Would it not work well if it was just `secondary="Total"`? (also below)

src/sections/TransactionInfoSection.tsx
```diff
@@ -41,10 +41,20 @@ interface ParamTypes {
...
 
+interface TransactionLite {
+  hash: string
+  blockHash: string
+  timestamp: number
+  inputs?: Input[]
+  outputs?: Output[]
+  gasAmount: number
+  gasPrice: string
+}
```
nop33, line 52: Isn't there a similar interface in alephium-js?

lf94: Yep, very similar, but the differences were causing problems. Feel free to suggest code changes using `TransactionLike`, `Transaction`, `UnconfirmedTransaction`, `ConfirmedTransaction`, but I've tried and it isn't trivial.

nop33: Isn't this exactly the same? https://github.com/alephium/alephium-js/blob/upgrade-schemas/api/api-explorer.ts#L132-L144

lf94: Maybe I had a brainfart moment because yep, my visual cortex is interpreting this as exactly the same. :rofl:

package.json
lf94: It will need to be updated no matter what because it relies on api-explorer.ts updates - I'll leave this comment open until that happens

lf94:
I've addressed the suggestions - will ping everyone when the backend updates are ready! Thank you for the great comments!

README.md
```diff
@@ -0,0 +1,37 @@
...
+- Lint and fix: `npm run lint:fix`
+
+See `package.json`'s `script` property for more options.
+
+**Note**: You will probably want to change the variables in `.env.placeholder`
+to use your local `explorer-backend`.
+
+## Notes
+
+This project uses [alephium-js](https://github.com/alephium/alephium-js).
```
nop33, line 35:
```suggestion
This project uses [@alephium/sdk](https://github.com/alephium/js-sdk).
```

README.md
```diff
@@ -0,0 +1,37 @@
...
+It's suggested to download `alephium-js` to refer to it as documentation on
```
nop33, line 36:
```suggestion
It's suggested to download it to refer to it as documentation on
```

nop33 (changes requested):

PR #64, opened 2022-03-23
UI cleaning

mvaivre:
Some quick cleaning and improvements :

- Reduce overall length to improve readability
- Reduce `bodyOnly` first column length
- Fix AppHeader (center it properly).
- AddressInfo: display TX list fetch error if any

### DIFF
--- a/src/sections/AddressInfoSection.tsx
+++ b/src/sections/AddressInfoSection.tsx
@@ -170,12 +170,12 @@ const TransactionInfoSection = () => {
           </>
         ) : (
           <TableBody>
-            <NoTxMessage>{txListError ?? 'No transactions yet'}</NoTxMessage>
+            <NoTxMessage>{txListError || 'No transactions yet'}</NoTxMessage>
           </TableBody>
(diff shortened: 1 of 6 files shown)


src/sections/AddressInfoSection.tsx
```diff
@@ -168,7 +170,7 @@ const TransactionInfoSection = () => {
           </>
         ) : (
           <TableBody>
-            <NoTxMessage>No transactions yet</NoTxMessage>
+            <NoTxMessage>{txListError ?? 'No transactions yet'}</NoTxMessage>
```
nop33, line 173: This will only return `No transactions yet` if `txListError` is specifically `undefined` or `null` ([docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator)). The initial value of `txListError` seems to be `''`. I think what you want here is `txListError || 'No transactions yet'`.

nop33 (approved):

PR #73, opened 2022-04-06
Fix statistics numbers

nop33:
## Before

image

## After

image


src/components/Statistics.tsx
```diff
@@ -33,16 +34,14 @@ interface Props {
 const Statistics = ({ refresh }: Props) => {
   const { client } = useGlobalContext()
   const [hashrate, setHashrate] = useState('')
-  const [totalSupply, setTotalSupply] = useState('')
-  const [circulating, setCirculating] = useState('')
-  const [transactions, setTransactions] = useState('')
-  const [blocks, setBlocks] = useState('')
-  const [avgBlockTime, setAvgBlockTime] = useState('')
+  const [totalSupply, setTotalSupply] = useState<number>()
+  const [circulating, setCirculating] = useState<number>()
```
lf94, line 38: Nit: I had kept all numbers as `string` to avoid precision loss as these numbers get higher.

src/utils/strings.ts
```diff
@@ -29,32 +29,44 @@ export const checkHexStringValidity = (stringToTest: string) => {
...
+  return (
+    timeSuffices.reduce((cur: string | undefined, [magnitude, suffix]) => {
+      if (cur !== undefined) return cur.toString()
+      if (n / magnitude < 1.0) return cur
+      return (n / magnitude).toPrecision(3) + suffix
+    }, undefined) || n.toString()
+  )
+}
+
+export const formatNumberForDisplay = (
```
lf94, line 48: Nit: `abbreviateValue` could have been adjusted to return `[string, string, string]` and additional suffix could have been added instead (it's a general function which will work with any measurement)

lf94 (approved):
Just two nits - otherwise everything is fine and can be merged! Thank you @nop33 :slightly_smiling_face:

src/components/Statistics.tsx
nop33: Since the numbers we receive from the API are integers and since these numbers won't ever reach the `MAX_SAFE_INTEGER` (9,007,199,254,740,991) (or will they?), then we can safely use `parseInt` to set the values in the state. Nothing actually changes, since you were also calling `parseInt` on the value stored in the state, before displaying them.

nop33:
Improved even further

image

mvaivre (approved):
:)

PR #74, opened 2022-04-06
Fix browser upgrade msg when BigInt not supported

nop33:
When running the app on an iPhone 11 Safari 13 through Browserstack, I noticed that our detection mechanism of whether `BigInt` is supported is broken:
```
ReferenceError: Can't find variable: BigInt
```

The reason is that we import the global context (which itself imports the sdk, which references `BigInt` in `lib/constants.ts`) BEFORE the BigInt detection mechanism.

This PR fixes this problem, importing the global context after the `BigInt` detection mechanism.

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -22,7 +22,6 @@ import ReactDOM from 'react-dom'
 import { HashRouter as Router } from 'react-router-dom'
 
 import NotificationBar from './components/NotificationBar'
-import { GlobalContextProvider } from './contexts/global'
 import * as serviceWorker from './serviceWorker'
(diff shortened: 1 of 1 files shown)


lf94:
Couldn't this be resolved by importing `App` at the top level instead of dynamically in the code?

The change itself looks good to me.

lf94 (approved):

nop33:
> Couldn't this be resolved by importing App at the top level instead of dynamically in the code?

`App` imports components that they themselves import the SDK. So no, it wouldn't work, that's why App is imported dynamically.

mvaivre (approved):

Issue #77, opened 2022-04-14
Differ stats loading


nop33:
Separate the loading of the stats so that once a stat is loaded, it's immediately displayed without waiting for the rest. For example, total transactions might take longer if it's not cached but all other stats take <100ms

PR #78, opened 2022-04-14
Improve loading stats

nop33:
As requested in #77 

https://user-images.githubusercontent.com/1579899/163345063-9b72c924-6329-432a-b157-6613eb3946f5.mp4

### DIFF
--- a/src/components/Statistics.tsx
+++ b/src/components/Statistics.tsx
@@ -38,19 +38,24 @@ interface Props {
 const Statistics = ({ refresh }: Props) => {
   const { client } = useGlobalContext()
   const [hashrate, setHashrate] = useState('')
+  const [isHashrateLoading, setIsHashrateLoading] = useState(true)
   const [totalSupply, setTotalSupply] = useState<number>()
+  const [isTotalSupplyLoading, setIsTotalSupplyLoading] = useState(true)
   const [circulating, setCirculating] = useState<number>()
+  const [isCirculatingLoading, setIsCirculatingLoading] = useState(true)
   const [transactions, setTransactions] = useState<number>()
+  const [isTransactionsLoading, setIsTransactionsLoading] = useState(true)
   const [blocks, setBlocks] = useState<number>()
+  const [isBlocksLoading, setIsBlocksLoading] = useState(true)
   const [avgBlockTime, setAvgBlockTime] = useState<number>()
-  const [isLoading, setIsLoading] = useState(true)
+  const [isAvgBlockTimeLoading, setIsAvgBlockTimeLoading] = useState(true)
 
   const currentSupplyPercentage = circulating && totalSupply && ((circulating / totalSupply) * 100).toPrecision(3)
 
   useEffect(() => {
     if (!client) return
 
-    const fetchData = async () => {
+    const fetchHashrateData = async () => {
       const now = new Date().getTime()
       const yesterday = now - ONE_DAY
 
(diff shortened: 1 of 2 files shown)


tdroxler:
:rocket:

tdroxler (approved):

src/components/Statistics.tsx
```diff
@@ -38,53 +38,83 @@ interface Props {
...
+  const [isTotalSupplyLoading, setIsTotalSupplyLoading] = useState(true)
   const [circulating, setCirculating] = useState<number>()
+  const [isCirculatingLoading, setIsCirculatingLoading] = useState(true)
   const [transactions, setTransactions] = useState<number>()
+  const [isTransactionsLoading, setIsTransactionsLoading] = useState(true)
   const [blocks, setBlocks] = useState<number>()
+  const [isBlocksLoading, setIsBlocksLoading] = useState(true)
   const [avgBlockTime, setAvgBlockTime] = useState<number>()
-  const [isLoading, setIsLoading] = useState(true)
+  const [isAvgBlockTimeLoading, setIsAvgBlockTimeLoading] = useState(true)
```
mvaivre, line 51: This starts to be a lot of individual states!
I wonder if we could maybe group the loading states together, like (pseudo code, not tested):

```ts
type StatKey = 'hashrate' | 'totalSupply' | ...

type LoadingState = { [key in StatsKey]: boolean }

// Initiate all loadings to true
const [loadingState, setLoadingState] = useState<>({...})

// Little util fn to merge previous and new states
const updateLoadingState = (newState: Partial<LoadingState>) => {setLoadingState((prev) => {...prev, newState}}

updateLoadingState({ hashrate: false })
```

mvaivre: The same technique could be used for storing and updating stats values.

```ts
type Stats = { [key in StatsKey]: number }

updateDataState({ hashrate: 42 })
```

src/components/Statistics.tsx
```diff
@@ -38,53 +38,83 @@ interface Props {
...
+
+    const fetchBlocksData = async () => {
       await client.infos
         .getInfosHeights()
         .then(({ data }) => setBlocks(data.reduce((acc: number, { value }) => acc + value, 0)))
 
+      setIsBlocksLoading(false)
+    }
+
+    const fetchTotalSupplyData = async () => {
```
mvaivre, line 77: Following the comment above, the next 3 fn could be a bit dried up 🌞 
Again, some pseudo code:

```ts
fetchData = async (call: () => Promise<HttpResponse>, statKey: StatKey) => {
  await call()
  .then((res) => res.text())
  .then((text) => updateDataState({[statKey]: parseInt(text)})
}
```

mvaivre (commented):
Some propositions to make the code dryer :) Tell me what you think!

src/components/Statistics.tsx
nop33: There 😛

src/components/Statistics.tsx
nop33: done

lf94: I would just combine it all then into a `State`. 

```
type State = { [k in StatKey]: { value: number | string; isLoading: boolean } }; const [state, setState] = useState<State>(defaultState);
```

nop33: I did that already

lf94 (approved):

PR #79, opened 2022-04-19
Animated counter

mvaivre:
Fun way to animate numbers between two states ✨ 
Ran on mount and between two renders for Blocks and Tx numbers.

### DIFF
--- a/src/components/Counter.tsx
+++ b/src/components/Counter.tsx
@@ -26,7 +26,7 @@ interface CounterProps {
 }
 
 const Counter = ({ from, to }: CounterProps) => {
-  const nodeRef = useRef<HTMLParagraphElement>(null)
+  const nodeRef = useRef<HTMLSpanElement>(null)
   const previousTargetValue = useRef(0)
 
   useEffect(() => {
--- a/src/components/Counter.tsx
+++ b/src/components/Counter.tsx
@@ -1,0 +1,51 @@
+/*
+Copyright 2018 - 2022 The Alephium Authors
+This file is part of the alephium project.
+
+The library is free software: you can redistribute it and/or modify
+it under the terms of the GNU Lesser General Public License as published by
+the Free Software Foundation, either version 3 of the License, or
+(at your option) any later version.
+
+The library is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+GNU Lesser General Public License for more details.
+
+You should have received a copy of the GNU Lesser General Public License
+along with the library. If not, see <http://www.gnu.org/licenses/>.
+*/
+
+import { addApostrophes } from '@alephium/sdk'
+import { animate } from 'framer-motion'
+import { useEffect, useRef } from 'react'
+
+interface CounterProps {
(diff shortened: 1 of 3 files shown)


src/components/Statistics.tsx
```diff
@@ -53,6 +54,8 @@ const Statistics = ({ refresh }: Props) => {
     avgBlockTime: statInitData
   })
 
+  console.log(statsData.totalTransactions)
```
lf94, line 57: Remove this line

src/components/Counter.tsx
```diff
@@ -0,0 +1,51 @@
...
+import { animate } from 'framer-motion'
+import { useEffect, useRef } from 'react'
+
+interface CounterProps {
+  from?: number
+  to: number
+}
+
+const Counter = ({ from, to }: CounterProps) => {
+  const nodeRef = useRef<HTMLParagraphElement>(null)
```
lf94, line 29: Should this be `HTMLSpanElement` or similar instead?

src/components/Counter.tsx
```diff
@@ -0,0 +1,51 @@
...
+      onUpdate(value) {
+        if (node) {
+          node.textContent = addApostrophes(value.toFixed(0))
+          previousTargetValue.current = to
+        }
+      }
+    })
+
+    return () => handle.stop()
+  }, [from, to])
```
lf94, line 46: `nodeRef`, should be in the list I think

lf94 (commented):
I like it!

src/components/Counter.tsx
mvaivre: Only states should be there I believe. Ref changes don't trigger renders. 
And the linter isn't screaming.

src/components/Counter.tsx
mvaivre: Yep, forgot to adjust after changing the element. Thanks!

src/components/Statistics.tsx
mvaivre: 🥱

src/components/Counter.tsx
lf94: > Only state should be there I believe

Nope: https://reactjs.org/docs/hooks-reference.html#conditionally-firing-an-effect

> all values from the component scope (such as props and state) that change over time and that are used by the effect.

It is possible `nodeRef` changes (the parent may cause the `Counter` to unmount and remount). This may be a Nit though, as the scenario may never happen depending on its usage.

mvaivre: The ref doesn't change during the lifecycle of a component. If my counter is unmounted and remounted by its parent, a new lifecycle starts. In the scope of the component, we get a new ref etc.
If you wanna manage refs from the parent perspective (for DOM measurements etc.), you can use `useCallback`.

Some links on the topic:

- https://epicreact.dev/why-you-shouldnt-put-refs-in-a-dependency-array/#:~:text=So%20the%20reason%20you%20shouldn,React%20when%20that%20change%20happens.
- https://stackoverflow.com/questions/60476155/is-it-safe-to-use-ref-current-as-useeffects-dependency-when-ref-points-to-a-dom
- https://medium.com/@teh_builder/ref-objects-inside-useeffect-hooks-eb7c15198780

nop33 (approved):

PR #88, opened 2022-04-27
Docker and docker-compose for a local development setup (full node, backend, db, frontend)

killerwhile:
This MR proposes a docker and docker-compose setup to run a local explorer (backend and frontend) connected to the testnet.


nop33:
omg, you're a hero 🦸 🚀

nop33:
Great job @killerwhile! I checked out your branch, started docker, and gave it a go:
```
cd dev
docker-compose up -d
```
I got an error:
```
ERROR: for alephium  Cannot start service alephium: Address already in use
```

Looking at my Docker app, I can see the `dev_frontend_1` as well as `dev_db_1` are running, but `dev_alephium_1` not:

image

When I manually stopped `dev_frontend_1` I could then start `dev_alephium_1` and then re-start `dev_frontend_1`:

image

--- 

Secondly, after I opened the frontend on my browser, I noticed that no data was loading. In the console I saw these errors:

```
Failed to load resource: net::ERR_CONNECTION_REFUSED

Uncaught (in promise) TypeError: Network request failed
    at XMLHttpRequest.S.s.onerror (browser-ponyfill.js:480:16)
```

Does that have anything to do with the networking with Docker?

---

Lastly, is there a specific reason to include your work in this repo? docker-compose starts the following services:
- Explorer frontend
- Explorer backend
- Alephium node

Maybe it'd be better fitted if we created a new repo (ie: `alephium-stack-docker`)? /cc @mvaivre @tdroxler @lf94 @polarker @h0ngcha0 @Lbqds @simerplaha opinions?

---

Thanks for your great work 👏

mvaivre:
> ERROR: for alephium  Cannot start service alephium: Address already in use

@nop33, did you make sure that your local node wasn't running before mounting the containers?

> Maybe it'd be better fitted if we created a new repo

Eventually, we may want to let people start the desktop wallet as well (web version) and maybe some of our dApps for testing purposes. So yes, creating a dedicated repo would make sense IMO. We would reference it in the node's / explorer backend's / frontend's repos.

nop33:
Yes

nop33:
> So yes, creating a dedicated repo would make sense IMO. We would reference it in the node's / explorer backend's / frontend's repos.

This project seems appropriate to host the docker stuff! https://github.com/alephium/alephium-stack

lf94:
I believe it's better for each project to be responsible for its own Dockerfile, but for composition, 100% agree that there should be a project which composes them all.

mvaivre:
(Also got the same error as @nop33 btw 🤷‍♂️ )

nop33:
Can we close this in favor of this now? https://github.com/alephium/alephium-stack
Is all work properly transferred to that repo>?

Issue #80, opened 2022-04-19
Load data only when browser tab is in view


mvaivre:
Avoid useless server loads by fetching only when browser tab is active.

lf94:
IIRC this is quite simple to do.

mvaivre:
Wanna do it @lf94? Feel free to assign yourself, would be great 🙌 👌

lf94:
* https://www.npmjs.com/package/@react-hook/timeout
* https://www.joshwcomeau.com/snippets/react-hooks/use-timeout/#explanation

Going to do a small experiment.

lf94:
I discovered we're already using a `useInterval` hook taken from somewhere (https://www.joshwcomeau.com/snippets/react-hooks/use-interval/ references Dan Abramov as the original author). The implementation still uses `setInterval` internally so it's not an option.

We'll use the hook.

PR #92, opened 2022-05-04
Prevent repetitive data requests when app is not visible

lf94:
Resolves https://github.com/alephium/explorer/issues/80

Confirmed using Wireshark.
Does not immediately re-request when visibility comes back; the interval is reset to 10 seconds.
Adding `isAppVisible` to the `useEffect` which called `getBlocks` can enable the immediate request behavior, but when users switch back and forth between the explorer and other tabs, it'll do a refresh each time. Keeping the delayed behavior is more resource efficient.


nop33 (approved):
Very clean fix 👏  I checked out the code and I can verify that it works (no requests when tab is not visible)

mvaivre (approved):
Great job! Small footprint, easy to understand 👌

Issue #45, opened 2022-01-11
Show the number of confirmations in the explorer


tappikone:
Feature request to explorer: Show the amount of confirmations next to the "Confirmed" text. Helps when troubleshooting issues with CEX deposits.

tdroxler:
:+1:
We will first show as confirmation the difference between the latest height of the given chain and the height of the block.

PR #94, opened 2022-05-04
Add confirmations for a transaction

lf94:
Resolves https://github.com/alephium/explorer/issues/45

### DIFF
--- a/src/components/Badge.tsx
+++ b/src/components/Badge.tsx
@@ -28,12 +28,16 @@ interface BadgeProps {
   className?: string
   amount?: string | bigint | undefined
   prefix?: string
+  inline?: boolean
   floatRight?: boolean
   minWidth?: number
 }
 
-let Badge = ({ content, className, amount, prefix, minWidth, floatRight = false }: BadgeProps) => (
-  <div className={className} style={{ float: floatRight ? 'right' : 'left', minWidth }}>
+let Badge = ({ content, className, amount, prefix, minWidth, inline = false, floatRight = false }: BadgeProps) => (
+  <div
+    className={className}
+    style={{ display: inline ? 'inline' : 'block', float: inline ? 'none' : floatRight ? 'right' : 'left', minWidth }}
+  >
     {prefix && <span>{prefix}</span>}
     {amount ? <Amount value={BigInt(amount)} /> : content}
   </div>
@@ -58,7 +62,7 @@ const getBadgeColor = (badgeType: BadgeType, theme: DefaultTheme) => {
       color = theme.textPrimary
       break
     case 'neutralHighlight':
-      backgroundColor = theme.bgSecondary
+      backgroundColor = theme.bgTertiary
       color = theme.textPrimary
       borderColor = theme.borderPrimary
   }
--- a/src/components/Badge.tsx
+++ b/src/components/Badge.tsx
@@ -59,7 +59,7 @@ const getBadgeColor = (badgeType: BadgeType, theme: DefaultTheme) => {
       break
     case 'neutral':
       backgroundColor = theme.bgTertiary
-      color = theme.textPrimary
+      color = theme.textSecondary
       break
     case 'neutralHighlight':
       backgroundColor = theme.bgTertiary
--- a/src/components/Badge.tsx
+++ b/src/components/Badge.tsx
@@ -33,11 +33,8 @@ interface BadgeProps {
   minWidth?: number
 }
 
-let Badge = ({ content, className, amount, prefix, minWidth, inline = false, floatRight = false }: BadgeProps) => (
-  <div
-    className={className}
-    style={{ display: inline ? 'inline' : 'block', float: inline ? 'none' : floatRight ? 'right' : 'left', minWidth }}
-  >
+let Badge = ({ content, className, amount, prefix }: BadgeProps) => (
+  <div className={className}>
     {prefix && <span>{prefix}</span>}
     {amount ? <Amount value={BigInt(amount)} /> : content}
   </div>
@@ -71,13 +68,16 @@ const getBadgeColor = (badgeType: BadgeType, theme: DefaultTheme) => {
 }
 
 Badge = styled(Badge)`
-  ${({ type, theme }) => {
+  ${({ type, inline = false, floatRight = false, minWidth, theme }) => {
     const { color, backgroundColor, borderColor } = getBadgeColor(type, theme)
 
     return css`
       color: ${color};
       background-color: ${backgroundColor};
       border: 1px solid ${borderColor};
+      display: ${inline ? 'inline' : 'block'};
+      float: ${inline ? 'none' : floatRight ? 'right' : 'left'};
+      min-width: ${minWidth ? minWidth + 'px' : 'auto'};
     `
   }}
 
@@ -85,7 +85,6 @@ Badge = styled(Badge)`
   padding: 5px 10px;
   border-radius: 4px;
   font-weight: 600;
-  float: left;
   white-space: nowrap;
 `
 
--- a/src/components/Links.tsx
+++ b/src/components/Links.tsx
@@ -18,7 +18,7 @@ along with the library. If not, see <http://www.gnu.org/licenses/>.
 
 import dayjs from 'dayjs'
 import { ExternalLink } from 'lucide-react'
-import React, { FC } from 'react'
+import { FC } from 'react'
 import { Link, LinkProps } from 'react-router-dom'
 import styled from 'styled-components'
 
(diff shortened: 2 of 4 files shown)


src/sections/TransactionInfoSection.tsx
```diff
@@ -98,15 +115,7 @@ const TransactionInfoSection = () => {
...
-                      <Badge
-                        type="plus"
-                        content={
-                          <span>
-                            <Check style={{ marginRight: 5 }} size={15} />
-                            Confirmed
-                          </span>
-                        }
-                      />
+                      <Badge type="plus" content={<span> {confirmations} Confirmations </span>} />
```
nop33, line 118:
```suggestion
                      <Badge
                        type="plus"
                        content={
                          <span>
                            {confirmations ?? '0'} {confirmations === 1 ? 'Confirmation' : 'Confirmations'}
                          </span>
                        }
                      />
```

mvaivre:
I used this PR to fix the missing uppercase in the `Total Value` label. While I was at it, I applied @nop33 's feedback and adapted the spacing a bit, hope it's fine :)

mvaivre:
I also displayed the block height (replaced the hash) and adapted the layout a bit. Inspired by etherscan.

image

src/components/Badge.tsx
```diff
@@ -28,12 +28,16 @@ interface BadgeProps {
...
   floatRight?: boolean
   minWidth?: number
 }
 
-let Badge = ({ content, className, amount, prefix, minWidth, floatRight = false }: BadgeProps) => (
-  <div className={className} style={{ float: floatRight ? 'right' : 'left', minWidth }}>
+let Badge = ({ content, className, amount, prefix, minWidth, inline = false, floatRight = false }: BadgeProps) => (
+  <div
+    className={className}
+    style={{ display: inline ? 'inline' : 'block', float: inline ? 'none' : floatRight ? 'right' : 'left', minWidth }}
```
nop33, line 39: This is getting quite long and ugly, how about moving it to the styled component definition below? We already use the `type` prop, we can also use the `inline` and `floatRight` there as well.

src/components/Links.tsx
```diff
@@ -31,15 +31,9 @@ interface TightLinkProps extends LinkProps {
...
-export const TightLinkStrict: React.FC<TightLinkProps> = ({ maxWidth, text, ...props }) => {
-  return (
-    <div style={{ maxWidth: maxWidth, display: 'flex', overflow: 'hidden' }}>
-      <StyledLink {...props} data-tip={text}>
-        {smartHash(text)}
-      </StyledLink>
-    </div>
-  )
-}
+export const SimpleLink: React.FC<LinkProps> = ({ children, ...props }) => (
```
nop33, line 34: I know it's not your change, but we already import `FC` so no need to use it from `React`.

src/components/Links.tsx
```diff
@@ -57,6 +51,16 @@ export const TightLink: React.FC<TightLinkProps> = ({ maxWidth, text, ...props }
   )
 }
 
+export const TightLinkStrict: React.FC<TightLinkProps> = ({ maxWidth, text, ...props }) => {
```
nop33, line 54: Same here

src/sections/TransactionInfoSection.tsx
```diff
@@ -79,6 +95,20 @@ const TransactionInfoSection = () => {
...
+    const chain = chainHeights.find(
+      (c: PerChainValue) => c.chainFrom === txBlock.chainFrom && c.chainTo === txBlock.chainTo
+    )
+
+    if (chain) {
+      const chainHeight = chain.value
+      confirmations = chainHeight - txBlock.height + 1
+    }
+  }
+
```
nop33, line 111: Let's put this in a utility function at the end of the file, so that we don't need to use comments to explain the code (like the `isTxConfirmed` one) and to improve the readability of the whole file.
```ts
const calculateConfirmations = (txBlock: BlockEntryLite, chainHeights: PerChainValue[]): number => {
  let confirmations = 0

  if (txBlock && chainHeights) {
    const chain = chainHeights.find(
      (c: PerChainValue) => c.chainFrom === txBlock.chainFrom && c.chainTo === txBlock.chainTo
    )

    if (chain) {
      const chainHeight = chain.value
      confirmations = chainHeight - txBlock.height + 1
    }
  }

  return confirmations
}
```

src/sections/TransactionInfoSection.tsx
```diff
@@ -115,22 +146,29 @@ const TransactionInfoSection = () => {
...
-                        text={txInfo.blockHash || ''}
-                        maxWidth="550px"
-                      />
+                      <SimpleLink to={`../blocks/${txInfo.blockHash || ''}`}>{txBlock?.height.toString()}</SimpleLink>
+                      <span data-tip="Number of blocks mined since" style={{ marginLeft: 10 }}>
+                        <Badge
+                          type="neutral"
+                          content={
+                            <span>
+                              {confirmations ?? '0'} {confirmations === 1 ? 'Confirmation' : 'Confirmations'}
```
nop33, line 166: With the above change that I propose, confirmations is always a number, so no need to check for undefined.

```suggestion
                              {confirmations} {confirmations === 1 ? 'Confirmation' : 'Confirmations'}
```

src/sections/TransactionInfoSection.tsx
```diff
@@ -115,22 +146,29 @@ const TransactionInfoSection = () => {
...
                   <TableRow>
-                    <td>Block Hash</td>
+                    <td>Block</td>
                     <td>
-                      <TightLink
-                        to={`../blocks/${txInfo.blockHash || ''}`}
-                        text={txInfo.blockHash || ''}
-                        maxWidth="550px"
-                      />
+                      <SimpleLink to={`../blocks/${txInfo.blockHash || ''}`}>{txBlock?.height.toString()}</SimpleLink>
```
nop33, line 160: I see that you changed the value to be the block height instead of the block hash. Then maybe we should also update the label of this entry from `<td>Block</td>` to `<td>Block Height</td>` to be more accurate?

mvaivre: I followed Etherscan's approach here, where the link redirects the user to the block page and the height serves as an ID. 
But as we have multiple chain, it's true that the height can't be considered as a unique identifier. To be perfectly correct, we should specify the Chain Index in the TX details as well. WDYT?

nop33: Ah, ok, didn't check Etherscan, I only noticed the change in the value but not a change in the label of the table row.

I don't know how important is to mention in which chain the block of the transaction exists, in the transaction page. It is already mentioned in the block page. As long as we link to the block page from the transaction page, it seems enough.

Unless of course you want to use block height AND chain id for unique id. Cause block height alone is not unique. Block hash is however.

nop33: Another idea:

image

mvaivre: What about 

image

src/components/Badge.tsx
mvaivre: Done!

src/components/Links.tsx
mvaivre: Done!

src/components/Links.tsx
mvaivre: Done!

src/sections/TransactionInfoSection.tsx
mvaivre: Done!

src/sections/TransactionInfoSection.tsx
mvaivre: Done!

src/sections/TransactionInfoSection.tsx
mvaivre: (Done)

src/components/Badge.tsx
```diff
@@ -71,21 +68,23 @@ const getBadgeColor = (badgeType: BadgeType, theme: DefaultTheme) => {
...
+  ${({ type, inline = false, floatRight = false, minWidth, theme }) => {
     const { color, backgroundColor, borderColor } = getBadgeColor(type, theme)
 
     return css`
       color: ${color};
       background-color: ${backgroundColor};
       border: 1px solid ${borderColor};
+      display: ${inline ? 'inline' : 'block'};
+      float: ${inline ? 'none' : floatRight ? 'right' : 'left'};
+      min-width: ${minWidth ? minWidth + 'px' : 'auto'};
```
nop33, line 80: when `minWidth` is `0`, then we want `auto`, correct?

nop33 (approved):

src/sections/TransactionInfoSection.tsx
lf94: Could be nice to come up with a better format to display the chain with the block hash in the future. Maybe the chain can be a subscript or superscript or something... having to hover for it doesn't feel great, especially since a block hash is a technical value and so knowing the chain is important.

mvaivre: All the block details are one click away. Let's see if users complain :)

PR #98, opened 2022-05-30
Feature/new design - part 2

mvaivre:
Supersedes #96

New :

- Firefox fixes according to @nop33 feedback
- CardWithChart
- NakedChart
- Full screen card
- Misc improvements


LeeAlephium:
https://github.com/alephium/explorer-backend/pull/239 hasn't been merged yet, so there are no new bindings for alephium/sdk, and of course it isn't available in production yet

src/components/Cards/CardWithChart.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+import { ComponentProps } from 'react'
+import styled from 'styled-components'
+
+import NakedChart from '../Charts/NakedChart'
+import Card from './Card'
+
+interface CardWithChartProps extends ComponentProps<typeof Card> {
+  chartSeries: ComponentProps<typeof NakedChart>['series']
+  chartColors: ComponentProps<typeof NakedChart>['colors']
```
LeeAlephium, line 27: While not explicitly mentioned anywhere, `ComponentProps` appears ideal for higher order components where it's not easy to know what components' properties will be, but the `CardWithChard` component specifically merges two components, where this isn't a problem...

With that reasoning, I would suggest using this:

```suggestion
interface CardWithChartProps extends CardProps {
  chartSeries: NakedChart['series']
  chartColors: NakedChart['colors']
```

src/components/Wave/WaveEntity.ts
```diff
@@ -0,0 +1,61 @@
...
+  private base: number
+
+  constructor(waveLength: number[], amplitude: number, base: number, gradientColors: string[]) {
+    this.waveLength = waveLength
+    this.gradientColors = gradientColors
+    this.amplitude = amplitude
+    this.base = base
+  }
+
+  public draw = (ctx: CanvasRenderingContext2D, width: number, height: number, frequency: number): void => {
```
LeeAlephium, line 32: It would be nice to list the original source of where this code was derived, but that's me

src/components/Wave/WaveEntity.ts
```diff
@@ -0,0 +1,61 @@
...
+    ctx.beginPath()
+    ctx.moveTo(0, height)
+    if (this.waveLength.length < 3) {
+      return
+    }
+
+    for (let i = 0; i < width; i++) {
```
LeeAlephium, line 39: This code should change in the future to be much more efficient when drawing the curves. Currently it's drawing each line as a pixel. I would place a "TODO" here:

```suggestion
    // TODO: Don't draw a line for each pixel of the curve (optimization)
    for (let i = 0; i < width; i++) {
```

src/components/Wave/WaveEntity.ts
```diff
@@ -0,0 +1,61 @@
...
+
+      ctx.lineTo(i * 2.5, height - (100 + this.base * 50) + wave1 * wave2 * wave3 * 80 * this.amplitude)
+    }
+    ctx.lineTo(width, height)
+
+    const gradient = ctx.createLinearGradient(0, height / 1.8, 0, height)
+
+    // Add three color stops
+    gradient.addColorStop(0, this.gradientColors[0])
+    gradient.addColorStop(1, this.gradientColors[1])
```
LeeAlephium, line 52: Is it three or two?

src/components/Wave/Waves.tsx
```diff
@@ -0,0 +1,75 @@
...
+*/
+
+import { useRef } from 'react'
+import styled from 'styled-components'
+
+import useAnimationFrame from '../../hooks/useAnimationFrame'
+import { useWindowSize } from '../../hooks/useWindowSize'
+import WaveEntity from './WaveEntity'
+
+type WaveName = 'frontWave' | 'backWave'
```
LeeAlephium, line 26:
```suggestion
type WavePosition = 'front' | 'back'
```

LeeAlephium: (Better would be nameless and just use indices)

src/utils/charts.ts
```diff
@@ -0,0 +1,61 @@
...
+        return formatToDay(new Date(_value))
+      }
+    }
+    return _value
+  }
+
+export const formatSeriesNumber = (type: YAxisType, value: number): string => {
+  // TODO: Special formatting (TX/s, etc)
+  if (type === 'tx') {
+    return value.toString()
```
LeeAlephium, line 52: I'm not sure we still want to just return the current value at the point being inspected for transactions? Did we want to improve this in a future PR?

LeeAlephium (commented):
Just a few small comments; feel free to apply or ignore as you see fit :)

src/components/Cards/CardWithChart.tsx
mvaivre: So, to make sure I understood properly, you would like me to export the props from `Card` and `NakedChart` and declare types like so?

```ts
import NakedChart, { NakedChartProps } from '../Charts/NakedChart'
import Card, { CardProps } from './Card'

interface CardWithChartProps extends CardProps {
  chartSeries: NakedChartProps['series']
  chartColors: NakedChartProps['colors']
  isLoading: boolean
}
```

LeeAlephium: Yea

src/utils/charts.ts
mvaivre: Fine for now, but I'll keep to todo :)

src/components/AppHeader.tsx
```diff
@@ -39,7 +43,7 @@ const AppHeader = ({ className }: AppHeaderProps) => {
           <Logo alt="alephium" src={theme.name === 'light' ? logoLightSrc : logoDarkSrc} />
         </StyledLogoLink>
       </HeaderSideContainer>
-      <StyledSearchBar />
+      {(pathname !== '/' || (width && width <= deviceSizes.mobile)) && <StyledSearchBar />}
```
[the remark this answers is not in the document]
mvaivre: image

src/pages/HomePage/HomePage.tsx
```diff
@@ -0,0 +1,316 @@
...
+  gap: 40px;
+  align-self: center;
+
+  @media ${deviceBreakPoints.mobile} {
+    flex-direction: column;
+    gap: 30px;
+  }
+`
+
+const Heading = styled.h2``
```
nop33, line 250: Isn't an `<h2>` semantic enough for a heading? :)

nop33 (commented):
image

Is that a dangerous `git add .`? 😝

src/pages/HomePage/HomePage.tsx
mvaivre: To be honest, those are simply the remanent objects of previously styled components that have been simplified. 
I kept them there just because... I don't really like how the "native" html elements appear compared to the styled components 🙃 

image

I know, it's not a very good reason but hey. I'll remove them.

nop33 (commented):
Amazing job @mvaivre 👏  It is so sick how you adapted the charts, such a nice idea!

A few things I noticed while testing:
- ✅ Firefox looks great now
- The chart modals do not close when pressing the ESC button (it does correctly close however when clicking outside of the box, or when clicking the "X" button 👍)
- The "Hashrate per day" chart is missing the unit of hashrate (GH/TH/etc). It should be enough if it was added on the header of the modal I suppose.
image

mvaivre:
Suggested changes are done @nop33. Thanks for the feedback!

mvaivre:
@LeeAlephium The tx/day history endpoint has been deployed on staging, and the hashrate history endpoint is already live on prod!
We can now fetch the real endpoints, test everything on staging (many things have changed in the backend), and finally deploy on prod 🚀

LeeAlephium:
@mvaivre should be good to go

LeeAlephium (approved):

src/utils/dates.ts
```diff
@@ -16,23 +16,14 @@ You should have received a copy of the GNU Lesser General Public License
...
-  flex-direction: column;
-`
-
-export default AddressesSection
+export const formatToYearMonthDay = (dt: Date): string => {
+  const day = dt.getDate().toString().padStart(2, '0')
+  const month = (dt.getMonth() + 1).toString().padStart(2, '0')
+  const year = dt.getFullYear()
+  return `${day}/${month}/${year}`
+}
```
nop33, line 29: Isn't the name of this function a bit confusing, based on the output it has?

nop33 (approved):

package.json
```diff
@@ -7,7 +7,7 @@
     "npm": ">=7.24.2"
   },
   "dependencies": {
-    "@alephium/sdk": "0.0.5",
+    "@alephium/sdk": "^0.0.9",
```
nop33, line 10: Didn't we say we wanted to fix the version and not use the `^`?

mvaivre: We did indeed

src/utils/dates.ts
mvaivre: Indeed

PR #106, opened 2022-06-16
Properly show unconfirmed transactions

### LINKED ISSUE #105: Unconfirmed txs not shown anymore
I notice that our latest explorer doesn't show the `Unconfirmed` transactions anymore.

The page only showed the loader, while the backend was responding: (hiding details for privacy reasons)
 
```
{
  "type": "Unconfirmed",
  "hash": "<tx_id>",
  "chainFrom": 0,
  "chainTo": 0,
  "inputs": [
    {
      "outputRef": {
        "hint": 1,
        "key": "<key>"
      },
      "unlockScript": "1234"
    }
  ],
  "outputs": [
    {
      "amount": "1",
      "address": "<address>"
    },
    {
      "amount": "1",
      "address": "<address>"
    }
  ],
  "gasAmount": 1,
  "gasPrice": "1"
}
```

mvaivre:
Fixes #105

### DIFF
--- a/src/pages/TransactionInfoPage.tsx
+++ b/src/pages/TransactionInfoPage.tsx
@@ -86,10 +86,10 @@ const TransactionInfoPage = () => {
         setTxInfoStatus(status)
         setTxInfoError(error.detail || error.message || 'Unknown error')
       }
-      setLoading(false)
     }
 
     fetchTransactionInfo()
+    setLoading(false)
   }, [client, id])
 
   // Initial fetch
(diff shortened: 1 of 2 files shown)


mvaivre:
Merging directly as it's a limited scope, can't do much harm + we need this fix.

src/pages/TransactionInfoPage.tsx
```diff
@@ -115,138 +115,131 @@ const TransactionInfoPage = () => {
...
+                  <td>Outputs</td>
                   <td>
-                    <b>Total Value</b>
-                  </td>
-                  <td>
-                    <Badge
-                      type="neutralHighlight"
-                      amount={outputs && outputs.reduce<bigint>((acc, o) => acc + BigInt(o.amount), BigInt(0))}
-                    />
+                    {txInfo.outputs
```
LeeAlephium, line 203: * It should be impossible for outputs to be absent from transaction information (nature of UTXO)
* `[] ? 'truthy' : 'falsey' === 'truthy'` will evaluate to `true`, so if `outputs` is an empty list nothing will appear - it'd probably be helpful to show "no outputs" instead but this situation is impossible in the first place because of the first point

src/pages/TransactionInfoPage.tsx
```diff
@@ -115,138 +115,131 @@ const TransactionInfoPage = () => {
...
+                </td>
+              </TableRow>
+              <TableRow>
+                <td>
+                  <b>Total Value</b>
+                </td>
+                <td>
+                  <Badge
+                    type="neutralHighlight"
+                    amount={outputs && outputs.reduce<bigint>((acc, o) => acc + BigInt(o.amount), BigInt(0))}
```
LeeAlephium, line 234: `BigInt(0)` could be shorter: `0n`

src/pages/TransactionInfoPage.tsx
```diff
@@ -115,138 +115,131 @@ const TransactionInfoPage = () => {
...
-                  <td>Status</td>
-                  {isTxConfirmed(txInfo) ? (
-                    <td>
+                  <td>Block</td>
+                  <td>
+                    <SimpleLink
+                      to={`../blocks/${txInfo.blockHash || ''}`}
+                      data-tip={`On chain ${txChain?.chainFrom} → ${txChain?.chainTo}`}
+                    >
+                      {txBlock?.height.toString()}
```
LeeAlephium, line 163: If `txBlock` is `undefined` the link will have no text

If you're certain this is not undefined then https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#non-null-assertion-operator is the operator to use, not the optional chaining operator, to satisfy the typechecker

PR #108, opened 2022-06-28
Implement 404 page

### LINKED ISSUE #107: Explorer route not showing 404 on invalid path
For example: https://explorer.alephium.org/#/xxx is a blank page. Is it a feature or a bug 😅

nop33:
Closes #107

Tested it locally and it works. May I deploy it on testnet?

image

image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -57,6 +57,12 @@
         "npm": ">=7.24.2"
       }
     },
+    "node_modules/@adobe/css-tools": {
+      "version": "4.0.1",
+      "resolved": "https://registry.npmjs.org/@adobe/css-tools/-/css-tools-4.0.1.tgz",
+      "integrity": "sha512-+u76oB43nOHrF4DDWRLWDCtci7f3QJoEBigemIdIeTi1ODqjx6Tad9NCVnPRwewWlKkVab5PlK8DCtPTyX7S8g==",
+      "dev": true
+    },
     "node_modules/@alephium/sdk": {
       "version": "0.0.9",
       "resolved": "https://registry.npmjs.org/@alephium/sdk/-/sdk-0.0.9.tgz",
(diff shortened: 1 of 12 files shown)


mvaivre (approved):

nop33:
I take the PR approval as approval to deploy it on testnet :) I'll do that, share it with the team, and if everything's good, I'll merge to master.

mvaivre:
Yes sir, that's what it is!

nop33:
Deployed on testnet. The link that @polarker reported on his issue #107 now shows a 404 page: https://testnet.alephium.org/#/xxx

There are a few things to improve, however:
1. Show a proper `404` HTTP status code instead of `200` when loading an invalid path. I believe this would be solved by changing the configuration on Amazon S3: https://tecadmin.net/solving-react-404-error/. We could try this on testnet (@killerwhile ?)
2. Make sure that omitting the `/#/` from the URL still results in a `404`.

_PS: The charts are not working because probably the backend is not deployed (@killerwhile ?)._

polarker:
I switched from BrowserRouter to HashRouter to fix an issue with Github pages: https://stackoverflow.com/a/71985764

Now we could switch back I think.

nop33:
> Also, we need to check that "hash" links in our current wallet are properly redirected.

@mvaivre with this fix we ensure that our apps will alls till work even after deploying this to production. We now have the time to properly update all our links without stress.

https://github.com/alephium/explorer/pull/108/commits/93cee3a42762d58bc0d4361f7dc310fe5a69b4da

mvaivre:
Testing on testnet: when trying to access https://testnet.alephium.org/blocks, I'm getting a white page.

nop33:
Benoit mentioned on Slack:

> The change is done. Error page is now index.html as well.

However, I still get a blank page when visiting https://testnet.alephium.org/blocks.

I followed the react-router-dom v5 docs about creating a 404 page:

And based on [other resources](https://jvdevlab.com/blog/react/404), the only change we had to make on Amazon S3 was the Error Document to be set to index.html...

Another [related article](https://codeutility.org/reactjs-blank-page-when-refreshing-in-react-router-stack-overflow/) suggests to add `"homepage": "/"` in the package.json to solve this issue, but that still didn't work.

Same problem exists when navigating from the homepage to the block page and then refreshing.

Researching this blank page on the internet I mostly get results for react-router-dom v6. Maybe it's worth spending time to upgrade?

mvaivre:
That's hard to debug :/ 

Yes, let's upgrade to `react-router-dom` v6 first. If I remember correctly, the got rid of the "<Switch>" mechanism entirely, so this may have an impact.

nop33:
The issue persists even after upgrading to v6. My intuition tells me that this is an Amazon S3 problem.

If you notice, when adding the`#/` in front of a subpath (example: `testnet.alephium.org/#/blocks`) and then you load the page, it works correctly (don’t mind the removal of the `#` , it is done by my code, not react-router itself). This means that the index.html file and the react js code are properly loaded. This is confirmed when looking into the network tab of the devtools.

When loading `testnet.alephium.org/blocks` however, the react code is not loaded. So I don’t think this is a react-router issue, rather a server config one.

- https://stackoverflow.com/questions/23543983/redirection-to-error-page-in-amazon-s3-static-website/23544903#23544903
- https://stackoverflow.com/questions/51218979/react-router-doesnt-work-in-aws-s3-bucket

nop33:
@mvaivre Ready for merge!

mvaivre (approved):

Issue #109, opened 2022-06-29
Significant CPU usage


LeeAlephium:
Doing nothing the explorer is using up 1/3rd of my processor activity. On refresh it's using up to 70%!

https://user-images.githubusercontent.com/105806603/176479356-f672e6d0-7e2d-4976-addd-7df35d24f6ff.mp4

nop33:
I think it's related to the "waves" on the footer. When I go to a subpage of the explorer that doesn't use the waves, the CPU usage drops significantly.

mvaivre:
Yeah as discussed with Lee, we can def optimize the rendering by avoiding drawing each pixel at each frame.
As a benchmark, It would be interesting to compare the cpu and gpu usage with another website using canvas animations.

LeeAlephium:
>  draw a polyline instead of multiple separate lines.

This should help the situation I'd think...Doubtful it'll remove the issue due to there still being a line per pixel

> If too complicated

Never :sunglasses: 

Let me know when you'd like me to spend time :rocket:

mvaivre:
Yes, your SVG idea sounds good to me @LeeAlephium ! 

Well, if you wanna spend an hour on it to see if this can be implemented, it would be great! Like that, we could release the website and an explorer update early next week.

PR #110, opened 2022-07-04
Optimise waves using less points and lower fps

LeeAlephium:
Brings the CPU usage down from ~70% peak and ~33% average to ~33% peak and ~7-10% average cpu usage

Resolves https://github.com/alephium/explorer/issues/109

### DIFF
--- a/src/components/Wave/WaveEntity.ts
+++ b/src/components/Wave/WaveEntity.ts
@@ -36,8 +36,11 @@ class WaveEntity {
       return
     }
 
-    // TODO: Don't draw a line for each pixel of the curve (optimization)
-    for (let i = 0; i < width; i++) {
+    const standardPpi = 96
+    const pointsPerInch = 15
+    const pointSpacing = standardPpi / pointsPerInch
+
+    for (let i = 0; i < width; i += pointSpacing) {
       const wave1 = Math.sin(i * this.waveLength[0] - frequency)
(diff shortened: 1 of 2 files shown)


LeeAlephium:
I tried heavy pre-computations for the wave calculations but even this had negligible effect. The bottleneck now I believe is the GPU process... Otherwise pre-computations would've done _something_.

LeeAlephium:
Using SVG is no better than the original way: it hits ~1/3rd of CPU usage as well, because it's probably optimizing for accuracy over performance (and thus probably also rendering per-pixel). The canvas based solution makes sense that it's faster: we're being very specific about our optimisation case.

LeeAlephium:
After doing a series of more tests, I believe the last 10% is just GPU<->CPU call overhead that just can't be overcome. When I basically have no paint calls it still remains at around 10%. I think then this is a good benchmark to have for future animations...

mvaivre:
Visually speaking, this looks like a good compromise to me ! 👏

mvaivre (approved):

mvaivre:
@LeeAlephium While you're at it, could you port your optimisation to the website 1.5? You can just add it to my latest [PR](https://github.com/alephium/www/pull/83). Thanks!

nop33 (approved):
I can confirm that this optimization has reduced the CPU usage considerably.

Issue #103, opened 2022-06-09
Broken SVG of the header logo in the dark theme


Sakrecoer:
As the title suggest, there is an issue with the SVG for the header logo on the dark theme:
![image](https://user-images.githubusercontent.com/3484593/172821217-cc62d2b2-efde-4dbf-a96e-dbf26c6deeec.png)

Works fine on the bright theme:
![image](https://user-images.githubusercontent.com/3484593/172821248-8c9947eb-6f78-483a-af8f-33bf8e7ff99e.png)

Issue #139, opened 2022-10-28
Use SVG Alef logo


mvaivre:
To avoid incorrect rendering in certain OSes

Issue #138, opened 2022-10-28
Fix amounts missaligments


mvaivre:
![image](https://user-images.githubusercontent.com/6715943/198553934-0364bf26-9056-4720-a530-f012740d8322.png)

Issue #136, opened 2022-10-24
QR code in address page isn't shown


mvaivre:
Let's change how we display the QR code in order to fix this.

If the screen is large, display the code by default, as in the mockups :

<img width="606" alt="image" src="https://user-images.githubusercontent.com/6715943/197585135-8fb32bd7-2616-442c-bc4f-7b32da235844.png">

If the screen is narrow, keep the current QR code button near the address hash, but display the code as a modal instead of a tooltip.

PR #140, opened 2022-10-28
Eject from CRA | Setup Vite.js | Misc UI improvements and fixes | Init UI for CSV Export

mvaivre:
Fixes :

#103
#112
#118 
#129 
#136 
#138 
#139 

Layout for #41


mvaivre:
I could hide the “Export CSV” button for now, so that we could ship this in prod.
Waiting for a first review before doing that.

src/components/PageSwitch.tsx
```diff
@@ -21,7 +21,8 @@ import { useCallback, useEffect, useMemo } from 'react'
 import { useLocation, useNavigate } from 'react-router-dom'
 import styled from 'styled-components'
 
-import useQueryParams from '../hooks/useQueryParams'
+import useQueryParams from '@/hooks/useQueryParams'
+
 import TextButton from './Buttons/TextButton'
```
nop33, line 26: I guess it's not important that some imports have not been updated, right? We can do that step by step in future PRs as we notice old-fashioned imports 😉

src/components/Select.tsx
```diff
@@ -0,0 +1,274 @@
...
+
+  const wrapperRef = useRef<HTMLDivElement>(null)
+  const listRef = useRef<HTMLDivElement>(null)
+
+  const { initialItemHeight, expandedItemHeight, initialListWidth, expandedListWidth } = dimensions as Required<
+    typeof dimensions
+  >
+
+  const orderedItems = [
+    ...items.filter(({ value }) => selectedItemValue === value),
```
nop33, line 77: I know this is almost silly optimization, but since you're only looking for 1 item here, this would be more efficient:

```ts
items.find(({ value }) => selectedItemValue === value)
```

src/components/Select.tsx
```diff
@@ -0,0 +1,274 @@
...
+  }
+
+  const handleClickOutside = () => {
+    setIsOpen(false)
+  }
+
+  const handleItemClick = (value: string) => {
+    setIsOpen(false)
+    onItemClick(value)
+  }
```
nop33, line 105: What I learnt recently is that any function that is not wrapped inside a `useCallback` (except the `setState` functions) will get redefined in every component re-render. This has the side-effect that if these functions are passed as props to other components, those components will re-render as well, even if they are memoized (since their props have changed).

So I started wrapping my callback functions that are passed as props with `useCallback`. This ensures that the children components props will change only when the depedencies of the callbacks change, which sounds correct to me.

src/components/Select.tsx
```diff
@@ -0,0 +1,274 @@
...
+
+  useOnOutsideClick(wrapperRef, handleClickOutside)
```
nop33, line 107: nit: How about we keep all the hooks at the top of the file, close to each other?

src/components/TextMiddleEllipsis.tsx
```diff
@@ -0,0 +1,87 @@
...
+    const visibleChars = Math.floor(el.current.clientWidth / charWidth.current)
+    const half = visibleChars / 2
+
+    setText(
+      visibleChars >= text.length
+        ? text
+        : text.slice(0, Math.floor(half)) +
+            (visibleChars == text.length ? '' : '...') +
+            text.slice(-Math.ceil(half) + 3)
+    )
```
nop33, line 76: I'm pretty sure this would have had the same effect without having to pass the state and set state functions are arguments. But I guess this was just copied from another project so maybe we can just add a comment to simplify it later, to not keep this PR stale.

```ts
  const handleResize = useCallback(() => {
    if (el?.current === null) return

    if (charWidth.current === undefined) {
      charWidth.current = el.current.scrollWidth / text.length
    }

    const visibleChars = Math.floor(el.current.clientWidth / charWidth.current)
    const half = visibleChars / 2

    setText(
      visibleChars >= text.length
        ? text
        : text.slice(0, Math.floor(half)) +
            (visibleChars == text.length ? '' : '...') +
            text.slice(-Math.ceil(half) + 3)
    )
  }, [text])

  useEffect(() => {
    handleResize()
  }, [handleResize])
```

src/hooks/useOnClickOutside.tsx
```diff
@@ -0,0 +1,37 @@
...
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+GNU Lesser General Public License for more details.
+
+You should have received a copy of the GNU Lesser General Public License
+along with the library. If not, see <http://www.gnu.org/licenses/>.
+*/
+
+import { RefObject, useEffect } from 'react'
+
+const useOnClickOutside = <T extends HTMLElement>(ref: RefObject<T>, handler: (e: MouseEvent) => void) => {
```
nop33, line 21: nit: It'd be nice if the hook arguments were defined in a `type` or `interface`

src/modals/ExportAddressTXsModal.tsx
```diff
@@ -0,0 +1,121 @@
...
+
+interface ExportAddressTXsModalProps extends ComponentProps<typeof Modal> {
+  addressHash: string
+}
+
+const ExportAddressTXsModal = ({ addressHash, ...props }: ExportAddressTXsModalProps) => {
+  const [timePeriodValue, setTimePeriodValue] = useState<TimePeriods>('7d')
+
+  return (
+    <Modal {...props} maxWidth={550}>
```
nop33, line 37: If we want components to be able to override the `maxWidth`, we'd need to move it before the `{...props}`. Otherwise all good.

src/modals/ExportAddressTXsModal.tsx
```diff
@@ -0,0 +1,121 @@
...
+      <FooterButton>
+        <Button accent big>
+          Export
+        </Button>
+      </FooterButton>
+    </Modal>
+  )
+}
+
+type TimePeriods = '7d' | '30d' | '6m' | '12m' | 'lastYear' | 'thisYear'
```
nop33, line 62: Shouldn't types be defined at the top of the file?

src/pages/AddressInfoPage.tsx
```diff
@@ -127,54 +135,59 @@ const TransactionInfoPage = () => {
...
+                ) : (
+                  <ErrorMessage>Could not get balance</ErrorMessage>
+                )}
+              </td>
+            </TableRow>
+            {totalBalance?.lockedBalance && parseInt(totalBalance.lockedBalance) > 0 && (
+              <TableRow>
+                <td>Locked Balance</td>
+                <td>
+                  <Badge type={'neutral'} amount={totalBalance.lockedBalance} />
```
nop33, line 176: `type="neutral"`

src/pages/AddressInfoPage.tsx
```diff
@@ -127,54 +135,59 @@ const TransactionInfoPage = () => {
...
-          )}
-        </TableBody>
-      </Table>
+            <TableRow>
+              <td>Total Balance</td>
+              <td>
+                {totalBalanceLoading ? (
+                  <LoadingSpinner size={14} />
+                ) : totalBalance ? (
+                  <Badge type={'neutralHighlight'} amount={totalBalance.balance} />
```
nop33, line 166: `type="neutralHighlight"`

src/utils/strings.ts
```diff
@@ -56,3 +56,4 @@ export const formatNumberForDisplay = (
 }
 
 export const DATE_TIME_FORMAT = 'DD/MM/YYYY HH:mm:ss [UTC]Z'
+export const SIMPLE_DATE_FORMAT = 'DD/MM/YYYY'
```
nop33, line 59: nit: How about merging these?

```suggestion
export const SIMPLE_DATE_FORMAT = 'DD/MM/YYYY'
export const DATE_TIME_FORMAT = `${SIMPLE_DATE_FORMAT} HH:mm:ss [UTC]Z`
```

nop33 (commented):
Done with the code review, I'll check out the code and run it now!

nop33:
Hm, the options seem invisible, yet clickable. On the dark theme it works fine. Moreover, switching themes doesn't re-render the button, only after hoverring over it.

image

image

image

nop33:
Hm, the QR code seems quite opaque in the light theme

image

src/components/PageSwitch.tsx
mvaivre: Exactly

src/components/Select.tsx
mvaivre: I'd love to do this, but I need to declare `handleClickOutside` first.

```Block-scoped variable 'handleClickOutside' used before its declaration.```

And this would be weird to declare the function before the hooks... no?

src/components/Select.tsx
mvaivre: Mhh this need to be evaluated (are the fns complex? Will the trigger a large component re-render?).

In that case I'm not so sure... Maybe the `onItemClick` callback that is received as a prop here should be using `useCallback`..

Some links (also shared in one of you PR):
- https://aheadcreative.co.uk/articles/when-to-use-react-usecallback/
- https://dmitripavlutin.com/react-usecallback/
- https://kentcdodds.com/blog/usememo-and-usecallback

src/modals/ExportAddressTXsModal.tsx
mvaivre: Done!

src/pages/AddressInfoPage.tsx
mvaivre: Done! Added a couple of corresponding eslint rules.

src/utils/strings.ts
mvaivre: Done!

mvaivre:
Color related issues with QRCode, Select and Button have been fixed.

src/hooks/useOnClickOutside.tsx
mvaivre: Done!

src/modals/ExportAddressTXsModal.tsx
mvaivre: Done!

src/components/Select.tsx
mvaivre: Done! (I had to cast the type though).

src/components/Select.tsx
nop33: Thanks for the articles, very insightful! Indeed, I see now that `useCallback` should be used with care.

What I am referring to is "[Referential equality](https://kentcdodds.com/blog/usememo-and-usecallback#referential-equality)". If a callback is passed from a parent component to a child component, and the child component uses this callback inside a `useEffect` or `useCallback`, then we need to ensure that that effect or callback will not run when the parent component simply re-renders because something inside of it changed, but not the prop callback itself.

src/components/TextMiddleEllipsis.tsx
nop33: I do. I am curius, however, what the future of js-sdk is. Now that we are moving parts of it to @alephium/web3, js-sdk might be a good candidate to host such components that are useful for frontend apps, since all it will contain is util functions for building frontends. What are your thoughts on this?

nop33 (approved):

src/components/TextMiddleEllipsis.tsx
nop33: I'm on my way to Neuch right now, hehe! But I'll be there next week as well.

I think I'd prefer a single repo, I don't see it as 180 turn, since js-sdk becomes more and more a "wallet/explorer"-sdk. But anyway, looking forward to discussing approaches next week!

I've approved the PR.

mvaivre: I've created a repo... juste in case :) For now, it's used as a placeholder for discussions. Cheers!

PR #145, opened 2022-11-10
Fix build & deploy

mvaivre:


nop33 (approved):
Let's fix the linting errors and it's good to go!

PR #152, opened 2022-12-12
Greatly simplify download handling

mvaivre:
Only disadvantage of this approach is that I can't specify the downloaded file name, because to the [`same-origin` requirement](https://stackoverflow.com/a/45368364).

But @tdroxler is now setting a default name in `1.11.1`, albeit a quite complex one with timestamps in millisecond. 

This should only be released after backend `1.11.1` is running on prod.

### DIFF
--- a/src/modals/ExportAddressTXsModal.tsx
+++ b/src/modals/ExportAddressTXsModal.tsx
@@ -42,55 +42,17 @@ const ExportAddressTXsModal = ({ addressHash, onClose, ...props }: ExportAddress
 
   const [timePeriodValue, setTimePeriodValue] = useState<TimePeriodValue>('24h')
 
-  const getCSVFile = useCallback(async () => {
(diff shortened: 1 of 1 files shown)


nop33 (approved):
Code looks good. Please, request another review once it's out of draft mode so that I can test it locally!

mvaivre:
After recent testing, it seems like the file extension isn't defined anymore. This probably has to be done in the backend @tdroxler 
I can't define the file name because of "same-origin" restrictions.

mvaivre:
@tdroxler added `.csv` to the filename and it's now working properly in every browser.
While we wait for the last explorer-backend ver to be released on prod, it's probably ok to ask our Chrome users to manually add the extension to their exported file (or to use Firefox or Safari).

nop33 (approved):

PR #153, opened 2022-12-12
Simplify table rows, allow middle click 

mvaivre:
Tables structure has been changed quite a bit. 
Need some focus on hands-on testing.

(I would have loved to clean up the styling which is a bit all over the place - between Table, TableRow etc.), but this would take too much time & effort, and it works as it is now. Will do next year ;))

### DIFF
--- a/src/components/Table/TableRow.tsx
+++ b/src/components/Table/TableRow.tsx
@@ -38,7 +38,9 @@ const TableRow: FC<RowProps> = ({ children, onClick, linkTo, className }) => (
     {Children.map(children, (c) =>
       linkTo ? (
         <td style={{ padding: 0 }}>
-          <FullHeightLink to={linkTo}>{c}</FullHeightLink>
+          <FullHeightLink className="row-link" to={linkTo}>
+            {c}
+          </FullHeightLink>
         </td>
       ) : (
         <td>{c}</td>
@@ -52,6 +54,10 @@ export default styled(TableRow)`
   border: none;
   cursor: ${({ onClick }) => (onClick ? 'pointer' : 'auto')};
   height: 45px;
+
+  td:first-child .row-link {
+    padding-left: 20px;
+  }
 `
 
 const FullHeightLink = styled(Link)`
@@ -61,10 +67,7 @@ const FullHeightLink = styled(Link)`
   display: flex;
   align-items: center;
   color: inherit;
-
-  &:first-child {
-    padding-left: 20px;
-  }
+  padding: 12px;
 
   &:hover {
     color: inherit;
--- a/src/components/Table/TableRow.tsx
+++ b/src/components/Table/TableRow.tsx
@@ -36,6 +36,7 @@ const rowVariants = {
 const TableRow: FC<RowProps> = ({ children, onClick, linkTo, className }) => (
   <motion.tr variants={rowVariants} transition={{ duration: 0.8 }} onMouseUp={onClick} className={className}>
     {Children.map(children, (c) =>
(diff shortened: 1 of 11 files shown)


src/components/Table/TableRow.tsx
```diff
@@ -31,14 +33,42 @@ const rowVariants = {
   shown: { opacity: 1 }
 }
 
-const TableRow: FC<RowProps> = ({ children, onClick, className }) => (
-  <motion.tr variants={rowVariants} transition={{ duration: 0.8 }} onClick={onClick} className={className}>
-    {children}
+const TableRow: FC<RowProps> = ({ children, onClick, linkTo, className }) => (
+  <motion.tr variants={rowVariants} transition={{ duration: 0.8 }} onMouseUp={onClick} className={className}>
+    {Children.map(children, (c) =>
```
nop33, line 38: Ha, I wasn't aware of the `Children` API! Looking at the docs I see that this API belongs to legacy and they recommend using alternatives: https://beta.reactjs.org/apis/react/Children

Maybe it's worth having a second look at this implementation. WDYT?

With a quick look, maybe something similar to how `headerTitles` is handled by the `TableHeader`?

src/pages/BlockInfoPage.tsx
```diff
@@ -209,28 +207,21 @@ const TransactionRow: FC<TransactionRowProps> = ({ transaction }) => {
...
-          <Badge
-            type="neutralHighlight"
-            amount={outputs?.reduce<bigint>((acc, o) => acc + BigInt(o.attoAlphAmount), BigInt(0))}
-            floatRight
-          />
-        </td>
+        </span>
+        <Badge
+          type="neutralHighlight"
+          amount={outputs?.reduce<bigint>((acc, o) => acc + BigInt(o.attoAlphAmount), BigInt(0))}
```
nop33, line 221: nit: I know this line was not created in this PR, but... 😅 I think the render function would look much nicer if this reducer was moved into a `totalAmount` variable.

src/pages/TransactionInfoPage.tsx
```diff
@@ -201,55 +195,47 @@ const TransactionInfoPage = () => {
...
-                    )}
-                  />
-                </td>
+                <b>Total Value</b>
+                <Badge
+                  type="neutralHighlight"
+                  amount={outputs?.reduce<bigint>(
+                    (acc, o) => acc + BigInt((o as Output).attoAlphAmount ?? (o as AssetOutput).attoAlphAmount),
+                    BigInt(0)
+                  )}
```
nop33, line 237: nit: 🤐 😅

nop33 (commented):
Wow, this whole change of table stuff to span stuff must have not been so much fun! Code review done, I'll do some functional testing now. 👏

nop33:
Functional tests look good, except only the press on the down/up arrow in both the Block and Address pages (both when opened or closed)


https://user-images.githubusercontent.com/1579899/207356410-88757799-d8c4-4c5b-a5f6-3432b75a6db6.mp4

src/components/Table/TableRow.tsx
mvaivre: Crap. I'll take a second look :)

mvaivre: TODO added

src/pages/BlockInfoPage.tsx
mvaivre: Done

mvaivre:
Everything's fixed + todo added :)

nop33 (approved):

PR #160, opened 2023-01-04
Fix multiple output display in tx table

nop33:
@mvaivre, this `Children.map` bit us in the 🍑 after all... When using the `TableRow` it's not obvious that its implementation will loop over the children and add each in a `td`... I guess that's why you removed the `width: 65%`, and now that I added it back it created this issue. Wrapping all outputs with a `div` fixes the problem in this case where only the first output is being displayed.


mvaivre:
Ah well...! We'll think about the table refactor when we'll want to add more feature to the explorer :) Thanks for the fix.

mvaivre (approved):

PR #164, opened 2023-02-02
Update testnet

nop33:


mvaivre (approved):

nop33:
hm, should we also update this in this PR? https://github.com/alephium/explorer/blob/master/.github/workflows/publish_to_mainnet.yml#L12

PR #165, opened 2023-03-02
Sdk update + Address Info grid + Asset list

mvaivre:


src/pages/AddressInfoPage/index.tsx
```diff
@@ -0,0 +1,325 @@
...
+    ...TokensMetadata[networkType].tokens.find((tm) => tm.id === t.id)
+  })) as Asset[]
+
+  if (totalBalance && lockedBalance && parseInt(totalBalance) > 0) {
+    assets?.push({ ...ALPH, balance: BigInt(totalBalance), lockedBalance: BigInt(lockedBalance) })
+  }
+
+  if (assets) {
+    assets = sortBy(assets, 'name')
+  }
```
mvaivre, line 164: Not so happy about this part (declaring conditional variables + bit of processing), but I didn't want to spend much more time on this and risking overengineering. If you have any proposition, I'm all ears 🤝

nop33: What about this?

```tsx
const assets = (addressData?.tokens.map((t) => ({
    ...t,
    balance: BigInt(t.balance),
    lockedBalance: BigInt(t.lockedBalance),
    ...TokensMetadata[networkType].tokens.find((tm) => tm.id === t.id)
})) ?? []) as Asset[]

if (totalBalance && lockedBalance && parseInt(totalBalance) > 0) {
  assets.push({ ...ALPH, balance: BigInt(totalBalance), lockedBalance: BigInt(lockedBalance) })
}

....

<AssetList assets={sortBy(assets, 'name')} isLoading={addressDataLoading} />
```

mvaivre: That's already something, thanks.

package.json
```diff
@@ -8,7 +8,8 @@
   },
   "homepage": "/",
   "dependencies": {
-    "@alephium/sdk": "0.5.0",
+    "@alephium/sdk": "^0.6.0-rc.4",
```
nop33, line 11: You can probably use 0.6.0 or 0.6.1

nop33: and I guess without the caret `^`

src/api/addressApi.ts
```diff
@@ -0,0 +1,63 @@
...
+): Promise<AddressTransactionsResult> => {
+  const { data: transactions } = await client.getAddressTransactions(addressHash, page)
+  const { data: unconfirmedTransactions } = await client.addresses.getAddressesAddressUnconfirmedTransactions(
+    addressHash
+  )
+
+  return {
+    addressHash,
+    transactions,
+    // Type casting needed due to https://github.com/alephium/explorer-backend/issues/427
```
nop33, line 60: Not needed anymore. You can use `MempoolTransaction`.

src/api/priceApi.ts
```diff
@@ -0,0 +1,27 @@
...
+*/
+
+type PriceResponse = { [tokenId: string]: { [currency: string]: number } }
+
+export const fetchAssetPrice = async (coinGeckoTokenId: string, currency = 'usd') => {
+  const res = (await (
+    await fetch(`https://api.coingecko.com/api/v3/simple/price?ids=${coinGeckoTokenId}&vs_currencies=${currency}`)
+  ).json()) as PriceResponse
+
+  return res[coinGeckoTokenId][currency]
```
nop33, line 26: Hm, what happens if for some reason the request fails (either because coingecko changes their API or just a simple network failure)? Accessing properties of an object that don't exist will lead to an error. Should we catch this error here, or somewhere else?

src/components/AssetLogo.tsx
```diff
@@ -0,0 +1,59 @@
...
+  justify-content: center;
+  align-items: center;
+  width: ${({ size }) => size}px;
+  height: ${({ size }) => size}px;
+  border-radius: ${({ size }) => size}px;
+  padding: 5px;
+  flex-shrink: 0;
+
+  ${({ asset, theme }) =>
+    asset.id === '0' // ALPH
```
nop33, line 47: In the desktop wallet, I created an object to identify the ALPH asset: https://github.com/alephium/desktop-wallet/blob/v2.0/src/storage/app-state/slices/assetsInfoSlice.ts#L28-L33

@h0ngcha0 came to me last week and asked me whether we should introduce such an object in the @alephium/token-list package and I think it's a great idea.

@h0ngcha0, did you move ahead with this, or should I take care of that?

@mvaivre do you agree?

EDIT: Just saw that you ported it from the desktop wallet later in this diff 😅

src/components/Buttons/TextButton.tsx
```diff
@@ -27,8 +27,8 @@ const StyledTextButton = styled.button`
   background: transparent;
   font-size: inherit;
   color: ${({ theme, disabled }) => {
-    if (disabled) return theme.textSecondary
-    else return theme.link
+    if (disabled) return theme.font.secondary
+    else return theme.global.accent
```
nop33, line 31: nit: I know you just changed the values, but let's fix this to:

```ts
color: ${({ theme, disabled }) => disabled ? theme.font.secondary : theme.global.accent};
```

src/components/Select.tsx
```diff
@@ -113,7 +113,6 @@ const Select = <T extends string>({
   return (
     <SelectWrapper
       role="button"
-      aria-label="Selected network"
```
nop33, line 116: No more accessibility on this? :)

src/components/Table/TableTabBar.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+interface TableTabBarProps {
+  items: TabItem[]
+  onTabChange: (tab: TabItem) => void
+  activeTab: TabItem
+  className?: string
+}
+
+const TableTabBar = ({ items, onTabChange, activeTab, className }: TableTabBarProps) => {
+  return (
```
nop33, line 34: nit: no need for `return` statement. Maybe we could update our format/linting rules to match the desktop wallet with this chance.

src/pages/AddressInfoPage/AddressTransactionRow.tsx
```diff
@@ -0,0 +1,192 @@
...
+            <TableHeader headerTitles={['Inputs', '', 'Outputs']} columnWidths={['', '50px', '']} />
+            <TableBody>
+              <TableRow>
+                <div>
+                  {t.inputs && t.inputs.length > 0 ? (
+                    t.inputs.map(
+                      (input, i) =>
+                        input.address && (
+                          <AddressLink
+                            key={i}
```
nop33, line 130: nit: the index is not a good key, maybe there's another unique value in the `input` object?

src/pages/AddressInfoPage/AddressTransactionRow.tsx
```diff
@@ -0,0 +1,192 @@
...
+const IconContainer = styled.div`
+  width: 28px;
+  height: 28px;
+  border-radius: 20px;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+`
+
+export default AddressTransactionRow
```
nop33, line 192: nit: could go higher up

src/pages/AddressInfoPage/InfoGrid.tsx
```diff
@@ -0,0 +1,102 @@
...
+    ) : (
+      <LoadingSpinner />
+    )}
+    {sublabel && <Sublabel>{sublabel}</Sublabel>}
+  </CellContainer>
+)
+
+// Subcomponent assignement
+
+InfoGrid.Cell = GridCell
```
nop33, line 58: Interesting approach!

src/pages/AddressInfoPage/index.tsx
```diff
@@ -0,0 +1,321 @@
...
+  // Asset price (appox).
+  // TODO: when listed tokens, add resp. prices. ALPH only for now.
+  useEffect(() => {
+    setAddressWorth(undefined)
+
+    const getAddressWorth = async () => {
+      const balance = addressData?.details.balance
+      if (!balance) return
+
+      const price = await fetchAssetPrice('alephium')
```
nop33, line 131: I guess it's here where we could use a try/catch to avoid network errors, similar to how we catch them above in this file

src/types/assets.ts
```diff
@@ -0,0 +1,33 @@
...
+GNU Lesser General Public License for more details.
+
+You should have received a copy of the GNU Lesser General Public License
+along with the library. If not, see <http://www.gnu.org/licenses/>.
+*/
+
+import { AddressBalance, Token } from '@alephium/sdk/api/explorer'
+import { TokenInfo } from '@alephium/token-list'
+
+// Used in Redux, amounts need to be in string format
```
nop33, line 22: 😅

src/utils/assets.ts
```diff
@@ -0,0 +1,26 @@
...
+*/
+
+import { NUM_OF_ZEROS_IN_QUINTILLION } from '@alephium/sdk'
+
+export const ALPH = {
+  id: '0',
+  name: 'Alephium',
+  symbol: 'ALPH',
+  decimals: NUM_OF_ZEROS_IN_QUINTILLION
+}
```
nop33, line 26: Ah, it's here as well!

nop33 (commented):
We need to use the throttled fetch from @alephium/web3, otherwise the explorer might not display the correct values due to the API rate limit. We can either do it in this PR or in another one I guess

src/components/AssetLogo.tsx
mvaivre: Yeah, I thought about this as well! I agree, this should be in the token-list package :)

src/api/priceApi.ts
mvaivre: I'll catch the error where this is used if that works for you (`AddressInfoPage/index`)

src/components/Select.tsx
mvaivre: Well, this was wrong to start with :P It's a generic `Select` component, not necessarily used for network selection.
The aria-label has to be set where the component is used (in `NetworkSwitch`). I'll do it :)

src/components/Table/TableTabBar.tsx
mvaivre: Done!

src/pages/AddressInfoPage/index.tsx
mvaivre: Didn't see your comment before, it's what I've done 💯

src/pages/AddressInfoPage/AddressTransactionRow.tsx
mvaivre: Used txHashRef

mvaivre:
Great feedback, thanks a lot! Everything has been fixed.

src/App.tsx
```diff
@@ -92,25 +92,23 @@ const App = () => {
...
-          >
-            {message.Icon}
-            {message.text}
-          </SnackbarPopup>
-        )}
-      </AnimatePresence>
-    </SnackbarManagerContainer>
-  )
-}
+const SnackbarManager = ({ message }: { message: SnackbarMessage | undefined }) => (
```
nop33, line 95: nit:

```suggestion
const SnackbarManager = ({ message }: { message?: SnackbarMessage }) => (
```

src/components/AlefSymbol.tsx
```diff
@@ -16,30 +16,32 @@ You should have received a copy of the GNU Lesser General Public License
...
+export default styled(AlefSymbol)`
+  height: 1em;
+`
 
 const HiddenForCopying = styled.span`
   font-size: 0;
 `
 
-const AlefSymbolSVG = styled.span`
+const AlefSymbolSVG = styled.span<{ color?: string }>`
```
nop33, line 43: The `color` prop is not used anymore, so it's not needed

nop33: From what I understand the `currentColor` property basically refers to the value of the `color` property. Since there is no `color` property defined here, how does this work?

src/components/Buttons/TextButton.tsx
nop33: I see a thumbs up but not a change, which is fine, it's a nit! Just curious why the thumbs up then 😅

nop33: ah, probably because you changed it in the second place of this file where it was needed.

nop33 (commented):
Love the new design of the Address page ✨ 

Probably not part of this PR, but when doing functional testing I noticed that all tooltips in the light theme are almost invisible:

image

Shouldn't the last activity be in human-readable format and the tooltip in UTC format?

image

Hm, aren't we educating our users wrongly when we list ALPH as a token? Technically it's a coin...

image

I checked Metamask and they indeed do it like we do it:

image

NIT: What is your opinion on aligning the titles and values of each block? Central vertical alignment looks a bit off to me

image

Not part of this PR probably, but I read this (2h) as "Average block time: 2 hours"!! To be frank, I don't even know what the 2h means here.

image

src/utils/assets.ts
nop33: You can use @alephium/token-list `v0.0.5` and get it from there!!

mvaivre:
Thanks for the comments. I'll check / fix this tomorrow!

src/components/AlefSymbol.tsx
mvaivre: It uses color inheritance if I'm not mistaken :)

mvaivre:
> Hm, aren't we educating our users wrongly when we list ALPH as a token? Technically it's a coin...

Yes, I spent some time thinking about this question. IMO, that's typically an exemple of useless technical complications making the crypto learning curve steeper. So yeah, I decided to use tokens as a generic term, I didn't find other options that would be simple enough, but you can see that I tried (cf. screenshot below :)). And yes, many other wallets seem to take the same shortcut. 

image


> NIT: What is your opinion on aligning the titles and values of each block? Central vertical alignment looks a bit off to me

I'm not shocked by this : many addresses won't have any locked coin + using "misalignment" kinda put the focus on the locked coins, which is good IMO. I'd say that it's good enough for a first iteration but I agree this could be more elegant. I'll revisit this once we want to show more info about "upcoming unlocking".

nop33 (changes requested):
```
Warning: Each child in a list should have a unique "key" prop.

Check the render method of `TokensList`.
```

Cannot load the address page. I am using `https://backend-v113.testnet.alephium.org` as the backend.

```
Uncaught TypeError: Cannot read properties of undefined (reading 'id')
    at AssetLogo.tsx:47:23
```

mvaivre:
As mentioned on Slack, I haven't been able to reproduce the couple issues you found. While I'm waiting for the "impacted" address hashes, I've improved the displayed of assets (esp. unknown assets) :)

nop33 (approved):
I cannot reproduce anymore ✅

PR #169, opened 2023-03-09
Ensure compatibility with explorer-backend v1.13.0

nop33:
@mvaivre I cherry-picked your commit and added 2 of my own. I know that your #165 PR was taking care of it, but I didn't want to rash merging something that might need more testing. This PR was done so that @killerwhile can retire the explorer-backend v1.11 endpoints which explorer mainnet is still using.

I've tested everything that I could think of locally using `https://backend-v113.testnet.alephium.org` as the backend URL and it works fine


mvaivre (approved):
LGTM! 🙏

PR #176, opened 2023-03-13
Fix last year CSV export

nop33:
The start datetime is set as the beginning of the day from 12 months ago and the end is set as the current datetime.

This would always result to more than exactly 1 year of duration (unless of course, this runs at exactly the beginning of the day at 00:00)

Removing the `startOf('day')` fixes the problem.

I also cleaned up the code a bit to make it more DRY.


mvaivre (approved):
LGTM! Thanks 🙏

PR #180, opened 2023-03-20
Assets in TXInfoPage & support showing of failed transactions

mvaivre:
Here's an example of failed TX : `91ed6771bba6ae5f1a39bac1b7d6ca3132fa49386c5af8ae8914061fac489560`

image

image


src/components/TransactionIOList.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+      if (io.tokens) {
+        amounts.push(...io.tokens.map((t) => ({ id: t.id, amount: BigInt(t.amount) })))
+      }
+
+      const IOAddressLink = (
+        <AddressLink
+          address={io.address}
+          txHashRef={(io as Input).txHashRef}
+          lockTime={(io as AssetOutput).lockTime}
```
nop33, line 47: this looks like a hack :/ The `io` cannot be both `Input` and `AssetOutput` at the same time, right?

I took the liberty to play a bit and find a different solution that only requires casting an `Output` as an `AssetOutput` (which is better than the current solution imo).

What do you think about the following implementation?

No need to spend too much time on this, I though this is an easy win for a cleaner code-base.

```tsx
interface TransactionIOListProps {
  inputs?: Input[]
  outputs?: Output[]
  flex?: boolean
  addressMaxWidth?: string
  IOItemWrapper?: ({ children }: { children: ReactNode }) => ReactElement
}

const TransactionIOList = ({
  inputs = [],
  outputs = [],
  flex,
  addressMaxWidth,
  IOItemWrapper
}: TransactionIOListProps) => {
  const getAmounts = (io: Input | Output) => [
    { id: ALPH.id, amount: BigInt(io.attoAlphAmount ?? 0) },
    ...(io.tokens ? io.tokens.map((t) => ({ id: t.id, amount: BigInt(t.amount) })) : [])
  ]

  const renderLink = (IOAddressLink: ReactNode, addressHash: string, index: number) =>
    IOItemWrapper !== undefined ? (
      <IOItemWrapper key={`${addressHash}-${index}`}>{IOAddressLink}</IOItemWrapper>
    ) : (
      IOAddressLink
    )

  return (
    <>
      {inputs.map((input, i) =>
        !input.address
          ? null
          : renderLink(
              <AddressLink
                address={input.address}
                txHashRef={input.txHashRef}
                amounts={getAmounts(input)}
                maxWidth={addressMaxWidth}
                flex={flex}
              />,
              input.address,
              i
            )
      )}
      {outputs.map((output, i) =>
        !output.address
          ? null
          : renderLink(
              <AddressLink
                address={output.address}
                lockTime={(output as AssetOutput).lockTime}
                amounts={getAmounts(output)}
                maxWidth={addressMaxWidth}
                flex={flex}
              />,
              output.address,
              i
            )
      )}
    </>
  )
}

export default TransactionIOList
```

src/utils/assets.ts
```diff
@@ -16,9 +16,29 @@ You should have received a copy of the GNU Lesser General Public License
...
+  networkType
+}: {
+  transaction: Transaction
+  addressHash: string
+  networkType: NetworkType
+}) => {
+  const { alph: alphAmount, tokens: tokenAmounts } = calcTxAmountsDeltaForAddress(transaction, addressHash)
+  const amount = convertToPositive(alphAmount)
+  const tokens = tokenAmounts.map((token) => ({ ...token, amount: convertToPositive(token.amount) }))
+  const tokenAssets = [...tokens.map((token) => ({ ...token, ...getAssetInfo({ assetId: token.id, networkType }) }))]
```
nop33, line 42: nit: instead of generating 3 arrays, this can be achieved with only 1 (but this is VERY NIT NIT NIT 😅)

```suggestion
const tokenAssets = tokenAmounts.map((token) => ({
    ...token,
    ...getAssetInfo({ assetId: token.id, networkType }),
    amount: convertToPositive(token.amount),
  }))
```

nop33 (commented):
Nice stuff man!

Not sure how to test the thing I see in the screenshot however. Everything seems to be working fine

src/components/HashEllipsed.tsx
```diff
@@ -41,13 +41,14 @@ const HashEllipsed = ({ hash, disableCopy = false, className, ...props }: HashEl
...
+  position: absolute;
+  right: 0;
   display: none;
-  height: 18px;
-  width: 18px;
   background-color: ${({ theme }) => theme.bg.accent};
   padding: 3px;
   border-radius: 4px;
+  min-height: 17px;
+  min-width: 17px;
```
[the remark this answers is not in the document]
mvaivre: Nice one. I had to refactor my component a lil' bit and I didn't use the exact same technique, but it works.

src/utils/assets.ts
mvaivre: I like it!

src/components/TransactionIOList.tsx
mvaivre: > The io cannot be both Input and AssetOutput at the same time, right?

Exactly, we know that for sure :) That's why I found my solution rather elegant. Taking a look a yours 👁️ ...

`Reading...`

I liked my solution TBH, mostly because it's DRYier and shorter. But I see your point, and I'm fine with your solution as well! 👍  Updated.

src/pages/TransactionInfoPage.tsx
```diff
@@ -132,16 +147,27 @@ const TransactionInfoPage = () => {
...
+                      }
+                      inline
+                    />
+                  ) : (
+                    <Badge
+                      type="minus"
+                      content={
+                        <>
+                          <span>Script execution failed</span>
+                        </>
```
nop33, line 167: Is the fragment needed? We can add this eslint rule:

```
'react/jsx-no-useless-fragment': 'warn',
```

nop33 (approved):

PR #181, opened 2023-03-24
Asset support (AssetList, tx lists and TX page)

mvaivre:
This combine my last approved PRs.

This can be used as a final check (manual testing) before releasing!


mvaivre:
Good catch. The issue was that the addresses' ellipsis wasn't resized as it should. It should now work with the current breakpoints.

nop33:
Most of the issue is solved, thanks! Just a tiny bit remaining though, we need more space allocated for the amount cause it can grow large.

image

I've added a commit to fix it. I'll leave it up to you to merge after you review it. You have my approval.

nop33 (approved):
Most of the issue is solved, thanks! Just a tiny bit remaining though, we need more space allocated for the amount cause it can grow large.

image

I've added a commit to fix it. I'll leave it up to you to merge after you review it. You have my approval.

PR #186, opened 2023-04-28
Fix last activity

mvaivre:
Fix last activity date and allow to manually set a page number to access to older items more easily.

### DIFF
--- a/src/components/PageSwitch.tsx
+++ b/src/components/PageSwitch.tsx
@@ -24,20 +24,19 @@ import styled from 'styled-components'
 
 import TextButton from '@/components/Buttons/TextButton'
 import useOnClickOutside from '@/hooks/useOnClickOutside'
-import useQueryParams from '@/hooks/useQueryParams'
+import usePageNumber from '@/hooks/usePageNumber'
 
 const manualInputHeight = 30
+const defaultElementsPerPage = 20
 
-const PageSwitch = ({
-  elementsPerPage = 20,
-  totalNumberOfElements,
-  numberOfElementsLoaded
-}: {
+interface PageSwitchProps {
+  totalNumberOfElements: number
   elementsPerPage?: number
-  totalNumberOfElements?: number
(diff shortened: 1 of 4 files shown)


src/components/PageSwitch.tsx
```diff
@@ -36,13 +40,36 @@ const PageSwitch = ({
...
+  useOnClickOutside({ ref: manualPageInputRef, handler: () => setIsSettingPageManually(false) })
+
   const handlePageSwitch = (direction: 'previous' | 'next') => {
     setPageNumber(direction === 'previous' ? currentPage - 1 : currentPage + 1)
   }
 
+  const handlePageInputChange = (e: ChangeEvent<HTMLInputElement>) => {
+    if (e.target.value === '') setManuallySetPage('')
+
+    const pageNumber = parseInt(e.target.value)
```
nop33, line 59: Hm, what if the result is `NaN` or some really huge number input that cannot be parsed?

src/pages/AddressInfoPage/index.tsx
```diff
@@ -104,8 +105,11 @@ const TransactionInfoPage = () => {
       setTxLoading(true)
       setAddressTransactions(undefined)
       try {
-        const transactionsData = await fetchAddressTransactions(client, id, pageNumber)
-        setAddressTransactions(transactionsData)
+        const firstPageTransactionData = await fetchAddressTransactions(client, id, 1)
+        const currentPageTransactionData = await fetchAddressTransactions(client, id, pageNumber)
```
nop33, line 109: If `pageNumber` is `1`, we are basically executing the same query twice, right?

nop33 (commented):
Nice job! 👏 

I found an input that breaks the pagination :D

`10000000000000002000`

http://127.0.0.1:5173/?p=10000000000000002000

image

mvaivre:
Let's say that it's a... nice easter egg? 🫠

nop33 (approved):

PR #187, opened 2023-05-01
Fix edge case of extremely large pagination

nop33:


src/pages/HomePage/HomePage.tsx
```diff
@@ -216,7 +222,7 @@ const HomePage = () => {
               </TableBody>
             </BlockListTable>
           </Content>
-          <PageSwitch totalNumberOfElements={blockList?.total} elementsPerPage={blockList?.blocks?.length} />
+          {blockList?.total && <PageSwitch totalNumberOfElements={blockList.total} elementsPerPage={8} />}
```
nop33, line 225: @mvaivre I purposfully set this to `8`. Without doing so, when we navigate to the last page there is the case that the length of the `blockList?.blocks` will be less than 8. Since the calculation of the `totalNumberOfPages` depends upon this number, I think it makes sense to always have it fixed.

src/components/PageSwitch.tsx
```diff
@@ -78,7 +77,8 @@ const PageSwitch = ({
     [locationSearch, navigate]
   )
 
-  const totalNumberOfPages = totalNumberOfElements && Math.ceil(totalNumberOfElements / elementsPerPage)
+  const totalNumberOfPages =
+    totalNumberOfElements && Math.ceil(totalNumberOfElements / (elementsPerPage || defaultElementsPerPage))
```
nop33, line 81: @mvaivre some times, `elementsPerPage` was `0`. Dividing by `0` caused the `Infinity`.

src/components/PageSwitch.tsx
```diff
@@ -91,6 +91,10 @@ const PageSwitch = ({
     }
   }, [currentPage, setPageNumber, totalNumberOfPages])
 
+  useEffect(() => {
+    setManuallySetPage(currentPage)
+  }, [currentPage])
```
nop33, line 96: @mvaivre this is to ensure that when the page is set through the URL, the same page is reflected when clicking to set it manually.

src/pages/HomePage/HomePage.tsx
nop33: Also, there is no reason to render the pagination switch unless we have `blockList?.total`

src/pages/HomePage/useBlockListData.ts
```diff
@@ -35,11 +36,16 @@ const useBlockListData = (currentPageNumber: number) => {
...
-      if (data) {
-        console.log('Number of block fetched: ' + data.blocks?.length)
-        setBlockList(data)
+        if (data) {
+          console.log('Number of block fetched: ' + data.blocks?.length)
+          setBlockList(data)
+          setPage(pageNumber)
+        }
+      } catch (e) {
+        setPage(1)
```
nop33, line 48: This was the only API query that was not wrapped in a `try/catch`, the rest of the pages work fine

src/components/PageSwitch.tsx
```diff
@@ -102,6 +106,7 @@ const PageSwitch = ({
       {isSettingPageManually ? (
         <PageManualInput
           step={1}
+          type="number"
```
nop33, line 109: What do you think about this? It should probably trigger the number pad on mobile. It also renders the up/down arrows on the browser.

src/components/PageSwitch.tsx
mvaivre: Mhh, I'm not sure that this is necessary?

src/components/PageSwitch.tsx
mvaivre: I started with adding the type, then removed it and forgot to remove the "step" as well.

When specifying `number` as a type, I couldn't prevent the input from showing letters (which doesn't make any sense IMO. Why allow a "number" input to display anything else than numbers..).

src/components/PageSwitch.tsx
nop33: What happens if you manuall a URL that has `?p=1000000000000000000000`, wait for the page to load, and then click the paginator? Without this effect, it will display `1000000000000000000000` instead of the actual number of the last page.

src/components/PageSwitch.tsx
nop33: ok, feel free to remove it, I'll focus on the windows thing today again...

src/components/PageSwitch.tsx
mvaivre: Ah I see! All clear now.

mvaivre (approved):

PR #193, opened 2023-05-10
Use stripped-down SDK

nop33:
Some stats are not showing up due to this: https://github.com/alephium/alephium-web3/issues/177

PR #200, opened 2023-05-23
Improve loading experience

mvaivre:
Wait for the new web3 version to be out and #193 to be merged.

### DIFF
--- a/src/api/addressApi.ts
+++ b/src/api/addressApi.ts
@@ -18,14 +18,43 @@ along with the library. If not, see <http://www.gnu.org/licenses/>.
 
 import { ExplorerProvider } from '@alephium/web3'
 
-import { AddressDataResult, AddressHash, AddressTransactionsResult } from '@/types/addresses'
+import {
+  AddressAssetsResult,
+  AddressBalanceResult,
+  AddressHash,
+  AddressTransactionNumberResult,
+  AddressTransactionsResult
+} from '@/types/addresses'
 
-export const fetchAddressData = async (
+export const fetchAddressBalance = async (
   client: ExplorerProvider,
   addressHash: AddressHash
-): Promise<AddressDataResult> => {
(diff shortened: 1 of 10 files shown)


src/api/addressApi.ts
```diff
@@ -18,14 +18,43 @@ along with the library. If not, see <http://www.gnu.org/licenses/>.
...
   addressHash: AddressHash
-): Promise<AddressDataResult> => {
+): Promise<AddressBalanceResult> => {
   const balances = await client.addresses.getAddressesAddressBalance(addressHash)
+
+  return {
+    addressHash,
+    balance: balances.balance,
+    lockedBalance: balances.lockedBalance
+  }
```
nop33, line 39: very nit:
```suggestion
  return {
    addressHash,
    ...balances
  }
```

On second thought, for something as simple as an API call, is creating a method in the api directory really needed? :/ In this particular case, we are simply wrapping the API call.

nop33: Same for below.

src/components/Cards/Card.tsx
```diff
@@ -28,17 +28,14 @@ export interface CardProps extends HTMLMotionProps<'div'> {
...
-  ) : (
-    <Container className={className} initial={false} {...props}>
-      <header>
-        <LabelText>{label}</LabelText>
-      </header>
-      <Content>{children}</Content>
-    </Container>
-  )
+const Card = ({ label, className, children, isLoading, ...props }: CardProps) => (
+  <Container className={className} initial={false} {...props}>
```
nop33, line 32: nit: the `className` could be removed so that it becomes part of `props`

nop33 (approved):
Some very nit comments, functional tests pass, approving anyways!

src/api/addressApi.ts
mvaivre: Done

PR #203, opened 2023-06-07
Mempool txs

mvaivre:


src/pages/AddressInfoPage/index.tsx
```diff
@@ -57,99 +59,139 @@ type ParamTypes = {
...
     balance: true,
     txNumber: true,
     assets: true,
     transactions: true
   })
 
-  const [addressWorth, setAddressWorth] = useState<number | undefined>(undefined)
-
-  const [exportModalShown, setExportModalShown] = useState(false)
+  const fetchTransactionNumber = useCallback(async () => {
```
mvaivre, line 89: @nop33 If you want a little challenge: dry those fetch calls bellow without overcomplexifying typings :))

src/contexts/global.tsx
```diff
@@ -26,7 +26,7 @@ import { OnOff } from '@/types/generics'
 import { NetworkType, networkTypes } from '@/types/network'
 import { SnackbarMessage } from '@/types/ui'
 
-interface GlobalContextInterface {
+export interface GlobalContextInterface {
```
nop33, line 29: nit: In our other apps we call these interfaces `[name]Props`. Maybe you'd like to change it now that you are exporting it?

src/pages/AddressInfoPage/AddressTransactionRow.tsx
```diff
@@ -16,10 +16,11 @@ You should have received a copy of the GNU Lesser General Public License
 along with the library. If not, see <http://www.gnu.org/licenses/>.
 */
 
-import { explorer } from '@alephium/web3'
+import { isMempoolTx } from '@alephium/sdk'
+import { MempoolTransaction, Transaction } from '@alephium/web3/dist/src/api/api-explorer'
```
nop33, line 20: Are you sure that this works well when building the app?

package.json
```diff
@@ -8,7 +8,7 @@
   },
   "homepage": "/",
   "dependencies": {
-    "@alephium/sdk": "0.7.0",
+    "@alephium/sdk": "v0.7.2",
```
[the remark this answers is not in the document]
mvaivre: 👀

src/contexts/global.tsx
mvaivre: Mhh good question! I associate props to components ("React props"). It kinda makes sense for me to use another naming scheme here.

src/pages/AddressInfoPage/AddressTransactionRow.tsx
mvaivre: I got no error when building.

src/contexts/global.tsx
nop33: Indeed, this TS interface does not represent props, but the **value** of the context. So, how about `GlobalContextValue`?

src/pages/AddressInfoPage/AddressTransactionRow.tsx
nop33: Hm, I would swear that there was a problem with importing like this and that's why I imported `explorer` and `node`...

nop33: maybe something changed in the web3 package, very weird. But good news :)

src/contexts/global.tsx
mvaivre: Why not!

src/pages/AddressInfoPage/AddressTransactionRow.tsx
```diff
@@ -111,72 +113,111 @@ const AddressTransactionRow: FC<AddressTransactionRowProps> = ({ transaction: t,
...
+            infoType === 'move'
+              ? 'Moved'
+              : infoType === 'out'
+              ? 'To'
+              : infoType === 'swap'
+              ? 'Swap'
+              : infoType === 'pending'
+              ? 'Pending'
+              : 'From'
+          }
```
nop33, line 126: nit: How about moving this into a variable instead?

```js
const badgeText = {
  move: 'Moved',
  out: 'To',
  swap: 'Swap',
  pending: 'Pending',
  in: 'From'
}[infoType]
```

even better, it would perfectly fit in `src/hooks/useTransactionUI.ts`.

src/pages/AddressInfoPage/AddressTransactionRow.tsx
```diff
@@ -111,72 +113,111 @@ const AddressTransactionRow: FC<AddressTransactionRowProps> = ({ transaction: t,
...
+const TableRowStyled = styled(TableRow)<{ pending: boolean }>`
+  ${({ pending, theme }) =>
+    pending &&
+    css`
+      > * {
+        opacity: 0.5;
+        animation: opacity-breathing 2s ease infinite;
+      }
+      background-color: ${theme.bg.secondary};
+      border-bottom: 1px solid ${theme.border.secondary};
```
nop33, line 205: super mega nit: I would give priority to the styles of the current component than to the styles of its children:

```suggestion
      background-color: ${theme.bg.secondary};
      border-bottom: 1px solid ${theme.border.secondary};
      
      > * {
        opacity: 0.5;
        animation: opacity-breathing 2s ease infinite;
      }
```

src/pages/AddressInfoPage/index.tsx
```diff
@@ -57,127 +59,106 @@ type ParamTypes = {
   id: string
 }
 
-const TransactionInfoPage = () => {
+type FetchedEntity = 'balance' | 'txNumber' | 'assets' | 'transactions'
```
nop33, line 62: nit: I have a few renames to suggest, cause I struggled parsing the code. How do you feel about renaming this to simply `AddressPropertyName`? (see below for other renames to make more sense).

src/pages/AddressInfoPage/index.tsx
```diff
@@ -57,127 +59,106 @@ type ParamTypes = {
...
-  const [loadings, setLoadings] = useState({
+  const [loadings, setLoadings] = useState<{ [key in FetchedEntity]: boolean }>({
     balance: true,
     txNumber: true,
     assets: true,
     transactions: true
   })
 
-  const [addressWorth, setAddressWorth] = useState<number | undefined>(undefined)
+  const fetchDataGeneric = useCallback(
```
nop33, line 93: nit: Continuing the renaming proposal, how about `fetchAddressDataGeneric`

src/pages/AddressInfoPage/index.tsx
```diff
@@ -57,127 +59,106 @@ type ParamTypes = {
...
+    async <T,>(
+      fetchedEntity: FetchedEntity,
```
nop33, line 95: nit: `addressPropName: AddressPropertyName`

src/pages/AddressInfoPage/index.tsx
```diff
@@ -57,127 +59,106 @@ type ParamTypes = {
...
+      setStateAction: (value: T | undefined) => void,
```
nop33, line 96: nit: `setAddressProp`

src/pages/AddressInfoPage/index.tsx
```diff
@@ -57,127 +59,106 @@ type ParamTypes = {
...
+      apiCallName?: SingleStringArgFunctions<ExplorerProvider['addresses']>,
+      customApiCall?: () => Promise<T>
```
nop33, line 98: nit: How about merging these 2 into one, since based on the implementation, only one of the 2 is used, even in the case if both are being provided? There is then no need to pass `undefined` as the third argument.

```suggestion
      apiCall: SingleStringArgFunctions<ExplorerProvider['addresses']> | (() => Promise<T>)
```

see below for completion

src/pages/AddressInfoPage/index.tsx
```diff
@@ -57,127 +59,106 @@ type ParamTypes = {
...
-
-    const fetch = async () => {
-      setLoadings((p) => ({ ...p, balance: true }))
-      setAddressBalance(undefined)
       try {
-        const balance = await client.addresses.getAddressesAddressBalance(id)
-        setAddressBalance(balance)
+        const result = customApiCall
+          ? await customApiCall()
+          : apiCallName && ((await client.addresses[apiCallName](id, {})) as T)
```
nop33, line 108: nit

```suggestion
const result = typeof apiCall === 'string' ? ((await client.addresses[apiCall](id, {})) as T) : await apiCall()
```

nop33 (commented):
👏  Small improvements that are not important to implement but worth mentioning:

1. Do not show a pointer on the pending tx row since it cannot be clicked to expand like the confirmed tx rows.
2. Clicking on the hash of a pending tx navigates to the tx page (which is great) but when going back, the pending tx is not listed and it only appears after a few seconds.

src/pages/AddressInfoPage/AddressTransactionRow.tsx
mvaivre: very good idea!

src/pages/AddressInfoPage/index.tsx
mvaivre: I like it! ❇️

nop33 (approved):
Extra nice ✨

PR #219, opened 2023-08-04
1.7.4 fixes

### LINKED ISSUE #217: [1.7.4] Planned improvements
- [ ] Port DW NFT's description overflow behaviour
- [x] AddessInfoPage: check whether the addressHash from useParams is a valid address before using it apart from whether it exists or not. Add the address validity check in the enabled flags of each query so that we avoid spamming the server with unnecessary requests when the address is invalid.
- [x] Remove axios from the deps (why the heck is it still here :P)
- [x] Create alphMetadata util object: `const alphMetadata = { ...ALPH, type: 'fungible', verified: true } as VerifiedFungibleTokenMetadata`
- [x] Amount.tsx: update suffix list based on sdk recent change
- [x] AssetLogo: handle potential failure of JSON.parse
- [x] staticDataContext: Make type more explicit: `<Map<Asset['id'], VerifiedFungibleTokenMetadata>>`
- [x] useTraceUpdate: remove es-lint disable

mvaivre:
Fixes #217

### DIFF
--- a/src/components/Amount.tsx
+++ b/src/components/Amount.tsx
@@ -16,7 +16,7 @@ You should have received a copy of the GNU Lesser General Public License
 along with the library. If not, see <http://www.gnu.org/licenses/>.
 */
 
-import { convertToPositive, formatAmountForDisplay, formatFiatAmountForDisplay } from '@alephium/sdk'
+import { MAGNITUDE_SYMBOL, convertToPositive, formatAmountForDisplay, formatFiatAmountForDisplay } from '@alephium/sdk'
 import styled from 'styled-components'
 
 import { useAssetMetadata } from '@/api/assets/assetsHooks'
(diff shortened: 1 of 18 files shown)


src/components/Amount.tsx
```diff
@@ -75,7 +75,7 @@ const Amount = ({
     if (value !== undefined) {
       amount = getAmount({ value, isFiat, decimals, nbOfDecimalsToShow, fullPrecision })
 
-      if (fadeDecimals && ['K', 'M', 'B', 'T'].some((char) => amount.endsWith(char))) {
+      if (fadeDecimals && MAGNITUDE_SYMBOL.some((char) => amount.endsWith(char))) {
```
nop33, line 78: Does the fact that `MAGNITUDE_SYMBOL` includes an empty string (`''`) affect this in any way?

nop33 (approved):

src/components/Amount.tsx
mvaivre: Such a great catch. I think it may affect this, indeed. We should probably export only the actual symbols from the js-sdk.

mvaivre: https://github.com/alephium/js-sdk/pull/158

mvaivre:
Just need to use sdk 0.7.5 & merge

PR #226, opened 2023-08-09
Use new grouped token balances fetch

mvaivre:


src/api/assets/assetsApi.ts
```diff
@@ -86,12 +86,28 @@ export const assetsQueries = createQueriesCollection({
...
   // TODO: This may be moved in a balancesApi file in the future?
   balances: {
-    addressToken: (addressHash: string, tokenId: string) => ({
-      queryKey: ['addressTokenBalance', addressHash, tokenId],
-      queryFn: () =>
-        client.explorer.addresses
-          .getAddressesAddressTokensTokenIdBalance(addressHash, tokenId)
-          .then((balance) => ({ ...balance, id: tokenId }))
+    addressTokens: (addressHash: string, tokenIds: string[]) => ({
+      queryKey: ['addressTokensBalance', addressHash, tokenIds.toString()],
```
nop33, line 90: why is `tokenIds` needed since they are not an input to the endpoint?

nop33 (commented):
As mentioned on Slack, I get too many NFTs for the `1DrDyTr9RpRsQnDnXo2YRiPzPW4ooHX5LLoqXrqfMrpQH` address whereas there should be only 4 (based on testnet swagger). Needs investigation

src/api/assets/assetsApi.ts
mvaivre: Haha, I was so sure you'd need to pass those that I didn't even double checked 🤦 
Much better than what I thought. Updating..

nop33 (approved):

PR #228, opened 2023-08-14
Upgrade react-tooltip

### LINKED ISSUE #215: Update the react-tooltip to v5 to allow using react 18
For now, we still hook our app using the `ReactDOM.render` which makes our apps function as if it was still using react 17.
Updating to the "new" way to render the app makes the tooltip package behave erratically.

nop33:
Closes #215


mvaivre (approved):
Awesome.

PR #229, opened 2023-08-14
Display more useful Total ALPH in TX details

mvaivre:
Just need to update the js-sdk version once out of rc.


src/pages/TransactionInfoPage.tsx
```diff
@@ -239,9 +240,33 @@ const TransactionInfoPage = () => {
                   fullPrecision
                 />
               </TableRow>
+              {tokensDeltaAmountsEntries.length > 0 && (
+                <TableRow>
+                  <b>Total Token Amounts</b>
+                  <AlphValuesContainer>
+                    {tokensDeltaAmountsEntries.map(([k, v]) => (
+                      <AlphValue key={k}>
+                        <AddressLink address={k} />
```
nop33, line 249: very nit: How about giving the key/value vars a more descriptive name, like `[addressHash, tokenAmounts]`?

src/pages/TransactionInfoPage.tsx
```diff
@@ -239,9 +240,33 @@ const TransactionInfoPage = () => {
...
+                    ))}
+                  </AlphValuesContainer>
+                </TableRow>
+              )}
               <TableRow>
-                <b>Total ALPH Value</b>
-                <Badge type="neutralHighlight" amount={totalAmount} />
+                <b>Total ALPH Amounts</b>
+                <AlphValuesContainer>
+                  {alphDeltaAmountsEntries.map(([k, v]) => (
```
nop33, line 263: same super nit here

src/styles/themes.ts
```diff
@@ -53,9 +53,9 @@ export const lightTheme: DefaultTheme = {
   global: {
     accent: '#5981f3',
     complementary: '#ce5cf8',
-    alert: '#ed4a34',
+    alert: '#ee5959',
     warning: '#ffa600',
-    valid: '#0ea68f',
+    valid: '#00a661',
```
nop33, line 58: Should we port these colors in our other apps as well?

src/styles/themes.ts
```diff
@@ -93,9 +93,9 @@ export const darkTheme: DefaultTheme = {
   global: {
     accent: '#598BED',
     complementary: '#eb88a4',
-    alert: '#ed4a34',
+    alert: '#f24242',
     warning: '#ffc42d',
-    valid: '#3ed9b0',
+    valid: '#1dcd84',
```
nop33, line 98: Same question

src/utils/transactions.ts
```diff
@@ -84,3 +84,86 @@ export const useTransactionInfo = (tx: Transaction | MempoolTransaction, address
     lockTime
   }
 }
+
+// TODO: The following 2 functions could be ported to js-sdk (and properly tested there)
```
nop33, line 88: I've created an issue: https://github.com/alephium/js-sdk/issues/160

src/utils/transactions.ts
```diff
@@ -84,3 +84,86 @@ export const useTransactionInfo = (tx: Transaction | MempoolTransaction, address
...
+type AttoAlphAmount = string
+type Address = string
```
nop33, line 90: Don't we already have a type for this called `AddressHash`?

src/utils/transactions.ts
```diff
@@ -84,3 +84,86 @@ export const useTransactionInfo = (tx: Transaction | MempoolTransaction, address
...
+
+type UTXO = {
+  attoAlphAmount?: AttoAlphAmount
+  address?: Address
+  tokens?: { id: string; amount: string }[]
```
nop33, line 95: We could use the `Token` type exported from explorer api.

src/utils/transactions.ts
nop33: I guess we only have it in the wallets

src/utils/transactions.ts
```diff
@@ -84,3 +84,86 @@ export const useTransactionInfo = (tx: Transaction | MempoolTransaction, address
...
+  return summed
+}
+
+export const sumUpTokenAmounts = (utxos: UTXO[]): Record<Address, Record<string, string>> => {
+  const validUtxos = utxos.filter((utxo) => utxo.address && utxo.tokens && utxo.tokens.length > 0)
+
+  const grouped = groupBy(validUtxos, 'address')
+  const summed = mapValues(grouped, (addressGroup) => {
+    const tokenSums: Record<string, string> = {}
+    for (const utxo of addressGroup) {
```
nop33, line 115: nit: space between var declaration and for loop

src/utils/transactions.ts
```diff
@@ -84,3 +84,86 @@ export const useTransactionInfo = (tx: Transaction | MempoolTransaction, address
...
+      for (const token of utxo.tokens || []) {
+        if (!tokenSums[token.id]) {
+          tokenSums[token.id]
+        }
```
nop33, line 119: I am not sure what this part of the code does. I would expect some sort of init assignment to happen here, like

```ts
tokenSums[token.id] = token.amount
```

src/utils/transactions.ts
```diff
@@ -84,3 +84,86 @@ export const useTransactionInfo = (tx: Transaction | MempoolTransaction, address
...
+        tokenSums[token.id] = (BigInt(tokenSums[token.id] || 0) + BigInt(token.amount)).toString()
+      }
```
nop33, line 121: While still trying to understand the algorithm I wonder if this implementation would be easier to grasp:

```ts
tokenSums[token.id] = !tokensSums[token.id] ? token.amount : (BigInt(tokenSums[token.id]) + BigInt(token.amount)).toString()
```


```

File: src/utils/transactions.ts [hunk 11]
```diff
@@ -84,3 +84,86 @@ export const useTransactionInfo = (tx: Transaction | MempoolTransaction, address
...
+    return tokenSums
+  })
+
+  return summed
+}
+
+export const IOAmountsDelta = (
+  inputs: UTXO[] = [],
+  outputs: UTXO[] = []
+): { alph: Record<string, string>; tokens: Record<string, Record<string, string>> } => {
```
nop33, line 132: nit: Since we have the `Address`, `AttoAlphAmount` and `Asset['id']` (from `web3`), wouldn't it be nicer to have better typing here than `string`?

src/utils/transactions.ts
```diff
@@ -84,3 +84,86 @@ export const useTransactionInfo = (tx: Transaction | MempoolTransaction, address
...
+      alphDeltas[address] = deltaAlph
+    }
+
+    const inputTokens = summedInputTokens[address] || {}
+    const outputTokens = summedOutputTokens[address] || {}
+    const allTokenIds = uniq([...Object.keys(inputTokens), ...Object.keys(outputTokens)])
+
+    allTokenIds.forEach((tokenId) => {
+      const deltaToken = (BigInt(outputTokens[tokenId] || 0) - BigInt(inputTokens[tokenId] || 0)).toString()
+      if (deltaToken !== '0') {
```
nop33, line 156: nit: Space between 😇

nop33 (commented):
👏 👏 👏 

Done with the code review! I'll do a functional test after you check my review!

src/styles/themes.ts
mvaivre: Not yet :) I'm making small adjustments here and there, as I feel like a theme refactor will be done in the future (more granular variables etc.).

src/utils/transactions.ts
mvaivre: Thanks a lot!

src/utils/transactions.ts
mvaivre: What happened here? 😆 Maybe just melted brain due to heat.

src/utils/transactions.ts
mvaivre: As said before, I think the condition is obsolete. This safe guard was needed at one point during my iterations, but based on the current logic and my recent testing it works well without it. To be tested further.

src/utils/transactions.ts
mvaivre: It would, definitely. I have to say that I was first focused to make it work quickly, but I'm now gladly adding more precise typings.

mvaivre:
Thanks for the review @nop33. There's one more thing I'd like to do (you'll see when doing the functional testing) : sorting the amounts, probably by asset name. It's not trivial though, because the name is fetched very late, when rendering the Amount component. 

If you have any proposition about sorting strategies and implementation, feel free to share! 🙏

nop33 (approved):
No more code comments to make ✅ 

image

Looking at this big transaction makes me wonder: Why do we need to separate the "Total Token Amounts" from the "Total ALPH Amounts" and not display only one "Total amounts" row with the first item always being "ALPH" (like we do in any transaction row in the tables in the wallets?

On the topic of UI: I really like how the transaction inputs and outputs are displayed in the table, next to each other. Did you by any chance consider such a UI for the deltas as well? 

image


Regarding sorting: Since the deltas are displaying the token symbol, why not sort by symbol? If we haven't yet fetched the symbol it doesn't make sense to display anything anyway.

nop33 (approved):
Nothing to add, great job! 👏

Issue #230, opened 2023-08-24
Show an error (and / or the 404 page) when a incorrect TX hash is passed in the URL


mvaivre:
We get an infinite loading if the TX hash is wrong, like: [/transactions/1HAgyKYN1McUxcDAP5Lv55EWHQJdR4ZUwrttt6VT5Qk5H](http://127.0.0.1:5173/transactions/1HAgyKYN1McUxcDAP5Lv55EWHQJdR4ZUwrttt6VT5Qk5H)

PR #232, opened 2023-08-28
Fix error handling in transaction page

tdroxler:
Resolves: https://github.com/alephium/explorer/issues/230

In case of a not found it was fine to keep retrying for 10 times to get the transaction as it could just be not yet synced by the explorer-backend or it's in between the mempool and the confirmed state.

If we receive any other error it's not worth retrying as it will always we wrong.

image
image
image

### DIFF
--- a/src/pages/TransactionInfoPage.tsx
+++ b/src/pages/TransactionInfoPage.tsx
@@ -16,7 +16,6 @@ You should have received a copy of the GNU Lesser General Public License
 along with the library. If not, see <http://www.gnu.org/licenses/>.
 */
 
-import { APIError } from '@alephium/sdk'
 import { ALPH } from '@alephium/token-list'
 import { explorer } from '@alephium/web3'
 import {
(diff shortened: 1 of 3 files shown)


src/pages/TransactionInfoPage.tsx
```diff
@@ -69,18 +71,24 @@ const TransactionInfoPage = () => {
   } = useQuery({
     ...queries.transactions.transaction.one(id || ''),
     enabled: !!id,
-    refetchInterval:
-      isAppVisible && (!previousTransactionData.current || !isTxConfirmed(previousTransactionData.current))
-        ? 10000
-        : undefined
+    retry: (num, e) => {
```
mvaivre, line 74: As discussed on the call:

- We need to re-add the refetchInterval (polling) logic
- We can export a numberOfAPIRetries variable from App.tsx

tdroxler: done

mvaivre (approved):

PR #236, opened 2023-09-12
FEAT: ADDED BAHASA INDONESIA TRANSLATION

ykris45:
Hello there. i am chris will added Bahasa Indonesia translations on Alephium Explorer. fell free to review my translations works here. Thank You so much for your assistance


mvaivre:
Thanks @ykris45 

I've used ChatGPT to try and validate the translations, you can find the output below. I'm sure your version is absolutely fine, but I thought sharing it with you may give you some ideas.

```
The translations provided are mostly accurate, with a few minor corrections and suggestions:

1. "Avg. block time": "Rata-rata" – It seems that the translation is incomplete. It should be "Rata-rata waktu blok".
2. "Nb. of assets": "Nb. dari aset" – "Nb." is short for "number" in English. In this context, a more accurate translation would be "Jumlah aset".
3. "Nb. of transactions": "Nb. dari transaksi" – Similarly, this would be better translated as "Jumlah transaksi".
4. "Hashrate per": "Hashrate per" – Without context, it's hard to provide a complete translation. If it refers to "Hashrate per block", then it should be "Hashrate per blok".
5. "Copy to clipboard": "Salin ke Papan" – A more commonly used translation would be "Salin ke papan klip".
6. "Fiat price": "Harga Mata uang" – A more precise translation would be "Harga dalam mata uang fiat".
7. "Number of blocks mined since": "Jumlah blok yang ditambang sejak itu" – The word "itu" (meaning "that") is unnecessary. It should be "Jumlah blok yang ditambang sejak".
8. "Our numbers": "Nomor kami" – Without context, this is tricky. If it refers to statistics or metrics, "Angka kami" might be more appropriate.

The rest of the translations appear to be correct. However, always keep in mind that context is crucial when translating phrases, especially in UI/UX design where space is limited and clarity is essential.
```

ykris45:
fixed @mvaivre

